PR #52, opened 2025-10-13
feat: Storage v2 binlog format read for MilvusDataSource

shaoting-huang:
related: https://github.com/zilliztech/spark-milvus/issues/74
Although milvus-storage provides an FFI interface that allows Spark to read and write binlogs in the Storage v2 format, the current Milvus 2.6 version uses the milvus-storage packed format, which does not include a manifest file.

To ensure compatibility with the Milvus 2.6 binlog format when using the FFI reader interface, we need to implement a manifest builder that generates a manifest file based on the segment information and Milvus schema.

This commit introduces a new Spark data source for reading Milvus Storage V2 binlog data. Key additions include:

- `MilvusDataSource`: Support Milvus 2.6+ binlog format read.
- `loon` module: handles manifest generation, and properties mapping.
-  FFI and JNI integration via `milvus-storage` to interface with native binlog parsing libraries.
- Utility packages (`schema`, etc.) to support schema mapping, and type conversion.
- Test suites covering manifest building, native integration, and source loading with Spark SQL logic.
- Updated `build.sbt` and added `spark_submit_demo.sh` for native linking and demo execution.

This enhancement enables Spark to efficiently read Milvus V2 binlog files and makes the connector compatible with the latest Milvus storage format.

Support ability:
- column projection
- pushed down filters
- vector search top K (For now: cosine, L2, inner product)
- Spark SQL


src/main/scala/MilvusDataReader.scala
```diff
@@ -15,18 +15,84 @@ object MilvusDataReader {
...
+        uri = config.uri,
+        token = config.token,
+        databaseName = config.options.getOrElse(MilvusOption.MilvusDatabaseName, "")
       )
-      .option(MilvusOption.ReaderType, "insert")
-      .option(MilvusOption.MilvusUri, config.uri)
-      .option(MilvusOption.MilvusToken, config.token)
-      .option(MilvusOption.MilvusCollectionName, config.collectionName)
-      .load()
+    )
```
liliu-z: Why we have these logic here. This class is just a sugar, it should not knwo anything about Milvus' internall concept like V1, V2. Let's hid this in a depper layer

src/main/scala/MilvusDataReader.scala
```diff
@@ -15,18 +15,84 @@ object MilvusDataReader {
...
+        .option(MilvusOption.MilvusToken, config.token)
+        .option(MilvusOption.MilvusCollectionName, config.collectionName)
+        .load()
+
+      v1DF.union(v2DF)
+    } else if (hasV2) {
+      // Only V2 segments
+      println(s"V2 collection detected, using storagev2 format")
+      spark.read
+        .format("storagev2")
```
liliu-z: We should not make this a different format. We only have one format registered, that is `milvus`

build.sbt
```diff
@@ -1,8 +1,30 @@
 import scala.sys.process.Process
+import scala.io.Source
 
 import xerial.sbt.Sonatype._
 import Dependencies._
 
+// Load Sonatype Central credentials
+credentials += {
```
liliu-z: Need to report the parsing failure

liliu-z (commented):
/lgtm
/approve

PR #62, opened 2025-12-11
perf: optimize V1 planInputPartitions() with batch API and eliminate S3 traversal

Thor-ChenBiao:
## Problem
  planInputPartitions() runs on Driver single-threaded. For N segments:
  - N HTTP calls to get segment info (one per segment)
  - 2N S3 listStatus calls to traverse binlog directories

  ## Solution

  ### 1. Batch Segment Info API
  - Add `getSegmentsInfoBatch()` in MilvusClient.scala
  - Single HTTP call fetches all segment insertLogIDs at once
  - N HTTP calls → 1 HTTP call

 ### 2. Eliminate S3 Traversal
  Before: Even though API returned insertLogIDs, code still traversed S3 to "verify":
  fs.listStatus(segmentPath)     // 1st S3 call: list field directories
    → fs.listStatus(fieldPath)   // 2nd S3 call: list binlog files per field
      → filter by insertLogIDs   // then filter results

  After: Trust API response, build paths directly via string concatenation:
  insertLogIDs: ["100/123456", "101/789012"]
                      ↓ split("/")
  fieldID: "100", logID: "123456"
                      ↓ concat
  fullPath: s"${rootPath}/${fieldID}/${logID}"

  Result: 2N S3 calls → 0 S3 calls

  ### 3. Parallel Segment Processing
  - Use .par.foreach with ConcurrentHashMap for thread-safe parallel processing
  - Note: After batch API optimization, this has minimal impact since remaining operations are pure memory/CPU (no I/O)

  ## Changes
  - MilvusClient.scala: Add getSegmentsInfoBatch() method
  - MilvusDataSource.scala:
    - Batch fetch at start of planInputPartitions()
    - New buildFieldMapFromLogInfo() builds paths without S3
    - Remove SegmentInfoCache (no longer needed with batch API)
    - Fail-fast on batch API failure (no silent fallback)

### DIFF
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -1,64 +1,25 @@
 package com.zilliz.spark.connector
 
-import java.io.File
-import java.net.http.{HttpClient, HttpRequest, HttpResponse}
-import java.net.URI
-import java.nio.charset.StandardCharsets
-import java.time.Duration
-import java.util.concurrent.TimeUnit
-import java.util.Base64
-import scala.concurrent.{ExecutionContext, Future, Promise}
-import scala.jdk.CollectionConverters._
-import scala.util.{Failure, Success, Try}
-
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.scala.{
-  DefaultScalaModule,
-  ScalaObjectMapper
-}
-import com.google.protobuf.ByteString
-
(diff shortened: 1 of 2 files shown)


LawrenceTL92:
Closing: this optimizes the Storage V1 (binlog) `planInputPartitions()` path, which was removed entirely in #69 (V2-only for Milvus 2.6+). Neither the S3 `listStatus` traversal nor `SegmentInfoCache` exist on main anymore, so there is nothing left for this change to apply to.

PR #68, opened 2026-01-12
fix: disable ASAN and fix ivy2 cache for Docker build

santiago-wjq:
- Disable ASAN (USE_ASAN=False) in native library build to avoid runtime compatibility issues
- Ensure conan profile and remote are configured in the same RUN command to handle empty cache
- Persist ivy2/local to workspace directory to fix COPY from BuildKit cache mounts
- Add Test/compile to build test classes


Thor-ChenBiao (approved):
LGTM

PR #73, opened 2026-01-16
fix: simplify snapshot manifest handling by passing basePath to datas…

shaoting-huang:
…ource

  - Change manifest handling to pass basePath instead of full manifest content, letting the FFI reader load manifests directly from storage
  - Fix S3 output path in BackfillConfig to exclude bucket name (configured separately via fs.bucket_name)
  - Add StorageV2ManifestItem companion object with factory method for Long segmentID serialization
  - Extract segmentID from manifest path in MilvusDataSource when not explicitly provided
  - Remove verbose logging statements throughout MilvusBackfill
  - Reduce test batchCount from 100 to 10 for faster test execution
  
  related: https://github.com/zilliztech/spark-milvus/issues/74


santiago-wjq (approved):
LGTM

PR #75, opened 2026-01-19
Fix: Propagate Hadoop S3 configuration to Executors

nianliuu:
This change fixes an issue where the Hadoop S3 configuration (specifically fs.s3a.aws.credentials.provider) was lost during serialization to executors because readerOptions was marked @transient lazy. It also improves logging and fallback logic for S3A credential providers.


santiago-wjq (approved):
LGTM

PR #76, opened 2026-02-12
enhance: support snapshot-only backfill with correct field IDs and fix data corruption

congqixia:
- Enable snapshot-only mode: extract collectionID, partition mapping, and segment base paths from snapshot manifests without requiring a live Milvus connection
- Write new fields with correct Milvus field IDs from snapshot schema; use numeric field ID as Arrow column name per milvus-storage convention
- Pin manifest reads to snapshot version (ver) for point-in-time consistency
- Fix UnsafeRow reuse corruption: .copy() before keyBy/partitionBy to prevent ExternalSorter from referencing the same mutable buffer
- Fix Arrow C Data zero-copy corruption: create new VectorSchemaRoot per flush, keep exported roots alive until C++ writer closes
- Use index-based Arrow vector lookup to support numeric field ID column names
- Return committedVersion from transaction commit; support ADDFIELD commit type
- Update milvus-storage submodule and align JNI JAR artifact name

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,6 +91,7 @@ lazy val root = (project in file("."))
   .settings(
     name := "spark-connector",
     assembly / parallelExecution := true,
+    assembly / assemblyPackageScala / assembleArtifact := false,
     Test / parallelExecution := true,
     Compile / compile / parallelExecution := true,
     version := s"${gitBranch}-${arch}-SNAPSHOT",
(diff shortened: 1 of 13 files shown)


liliu-z (dismissed):

liliu-z (changes requested):

liliu-z (changes requested):
-

liliu-z (commented):
/lgtm
/approve

PR #78, opened 2026-04-09
feat: support running on Spark Operator with IAM/IRSA and dual-bucket credentials

congqixia:
Make MilvusBackfill deployable as a standalone spark-submit job on Kubernetes (Apache Spark Operator), with first-class support for IRSA auth and an input parquet bucket separate from the Milvus storage bucket.

New entry point and config:
- BackfillApp: new spark-submit main class. Parses --parquet/--snapshot /--s3-* / --use-iam / --source-s3-* / --source-use-iam, builds a BackfillConfig and invokes MilvusBackfill.run(). Auto-enables use_iam when both ak/sk are empty so no flag is needed under IRSA.
- BackfillConfig: add s3UseIam plus optional sourceS3Endpoint / sourceS3AccessKey / sourceS3SecretKey / sourceS3UseSSL / sourceS3UseIam / sourceS3Region. Both read and write option builders forward fs.use_iam to the Milvus FFI so it picks up the AWS default credentials chain (env / web identity / instance profile).

Backfill core:
- MilvusBackfill: program per-bucket fs.s3a.bucket.<name>.* on the SparkContext hadoopConfiguration for both the snapshot path and the backfill parquet path, so each bucket can independently use static AK/SK or IAM. Stop unconditionally overwriting fs.s3a.access.key/ secret.key (which previously broke DefaultAWSCredentialsProviderChain under IRSA). Use the fully qualified MilvusDataSource class in spark.read.format(...) to avoid shortName collisions when other connectors also register "milvus".

Build / packaging:
- Mark hadoop-common, hadoop-aws and parquet-hadoop as provided,test — Spark images already ship them, and bundling caused "X not a subtype of Y" classloader errors (especially with userClassPathFirst=true).
- Globally exclude slf4j-log4j12 / slf4j-reload4j / log4j 1.x / reload4j to remove the duplicate StaticLoggerBinder against Spark's log4j2.
- Filter slf4j-api out of the assembly fat jar so the runtime org.slf4j.Logger only comes from /opt/spark/jars, eliminating LinkageError on the Logging trait when the user JAR is loaded first.

Submodule:
- Bump milvus-storage to pick up the JNI fat-jar SONAME-versioned .so packaging fix and the aws-c-* shared-lib build.

### DIFF
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -101,6 +101,37 @@ object BackfillApp {
     }
   }
 
+  // Boolean flags do not consume the next token.
+  private[backfill] val BoolFlags: Set[String] = Set(
+    "s3-use-ssl",
+    "use-iam",
+    "source-s3-use-ssl",
+    "source-use-iam"
+  )
+
+  // All accepted CLI keys. Validating against this whitelist surfaces typos
+  // (e.g. `--s3access-key` instead of `--s3-access-key`) at parse time
+  // instead of silently absorbing them and producing a confusing failure
+  // later inside the AWS default provider chain.
+  private[backfill] val KvFlags: Set[String] = Set(
+    "parquet",
+    "snapshot",
+    "s3-endpoint",
+    "s3-bucket",
+    "s3-access-key",
+    "s3-secret-key",
+    "s3-root-path",
+    "s3-region",
+    "source-s3-endpoint",
+    "source-s3-access-key",
+    "source-s3-secret-key",
+    "source-s3-region",
+    "batch-size",
+    "output-result"
+  )
+
+  private[backfill] val KnownFlags: Set[String] = BoolFlags ++ KvFlags
+
   private[backfill] def parseArgs(args: Array[String]): Map[String, String] = {
     var map = Map.empty[String, String]
     var i = 0
@@ -108,9 +139,14 @@ object BackfillApp {
       args(i) match {
         case flag if flag.startsWith("--") =>
           val key = flag.stripPrefix("--")
-          if (
-            key == "s3-use-ssl" || key == "use-iam" || key == "source-s3-use-ssl" || key == "source-use-iam"
-          ) {
+          if (!KnownFlags.contains(key)) {
+            throw new IllegalArgumentException(
+              s"Unknown argument: $flag. Known flags: ${KnownFlags.toSeq.sorted
+                  .map("--" + _)
+                  .mkString(", ")}"
+            )
+          }
+          if (BoolFlags.contains(key)) {
             map += (key -> "true")
             i += 1
           } else if (i + 1 < args.length) {
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -1,0 +1,120 @@
+package com.zilliz.spark.connector.operations.backfill
+
+import org.apache.spark.sql.SparkSession
+
+/** Spark application entry point for running backfill via spark-submit.
+  *
+  * Usage: spark-submit --class
+  * com.zilliz.spark.connector.operations.backfill.BackfillApp \
+  * spark-connector-assembly.jar \ --parquet <path> --snapshot <path> \
+  * --s3-endpoint <endpoint> --s3-bucket <bucket> \ --s3-access-key <key>
+  * --s3-secret-key <secret> \ [--s3-root-path <path>] [--s3-region <region>]
+  * [--s3-use-ssl] \ [--batch-size <n>] [--output-result <path>]
+  */
+object BackfillApp {
+
+  def main(args: Array[String]): Unit = {
+    val parsed = parseArgs(args)
+
+    val parquetPath = parsed.getOrElse(
+      "parquet",
+      throw new IllegalArgumentException("--parquet is required")
+    )
+    val snapshotPath = parsed.getOrElse(
+      "snapshot",
+      throw new IllegalArgumentException("--snapshot is required")
+    )
+    val s3Endpoint = parsed.getOrElse(
+      "s3-endpoint",
+      throw new IllegalArgumentException("--s3-endpoint is required")
+    )
+    val s3Bucket = parsed.getOrElse(
+      "s3-bucket",
+      throw new IllegalArgumentException("--s3-bucket is required")
+    )
+    // Optional in IAM/IRSA mode — when both empty, automatically enable use_iam so that
+    // both Milvus FFI and Spark Hadoop S3A defer to the default credentials chain
+    // (env vars, instance profile, web identity token, etc.).
+    val s3AccessKey = parsed.getOrElse("s3-access-key", "")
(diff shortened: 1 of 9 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -674,47 +822,147 @@ object MilvusBackfill {
...
-        out.write(result.toJson.getBytes(java.nio.charset.StandardCharsets.UTF_8))
+        out.write(
+          result.toJson.getBytes(java.nio.charset.StandardCharsets.UTF_8)
+        )
       } finally {
         out.close()
       }
       logger.info(s"Backfill result JSON written to: $outputPath")
     } catch {
       case e: Exception =>
```
liliu-z, line 893: writeResultJson swallows exceptions, causing silent success: exit 0 with success message printed but no result file written. Change return type to Either (or throw) and gate the success message in BackfillApp.scala:85 on the actual outcome.

liliu-z, line 893: writeResultJson has no access to per-bucket S3 configuration for its output path. Restructure the signature to pass config through.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -1,23 +1,36 @@
 package com.zilliz.spark.connector.operations.backfill
```
liliu-z, line 1: validate() must enforce `s3UseIam || (AK && SK)` as a hard invariant for both main and source bucket credentials.

liliu-z, line 1: sourceS3Region is exposed as a CLI parameter but never wired into configureHadoopS3ForPath (e.g. fs.s3a.bucket.<b>.endpoint.region). Either wire it through or remove it from the CLI/config.

liliu-z:
**src/test/scala/operations/backfill/BackfillConfigTest.scala:79** Assertions at lines 79 and 92 conflict with the new semantics — the test suite cannot be passing. Sync tests with the new behavior.

liliu-z:
**src/main/scala/operations/backfill/README.md:45** README is stale: references removed `pkFieldToRead` (lines 45, 87, 190, 227, 262), the old 3-arg `run()` signature (61-65), non-existent `MilvusBackfill.backfill()` (235), and `BackfillConfig.forTest` with pkFieldToRead (260-263). No mention of BackfillApp, --parquet, --snapshot, IAM/IRSA, --source-s3-*, dual-bucket, or the new spark-submit entry point. Must be rewritten in this PR.

liliu-z:
**src/test/scala/operations/backfill/BackfillConfigTest.scala:1** Missing tests for parseArgs, the new validate() invariants, configureHadoopS3ForPath Hadoop key writes, and an FQCN format smoke test.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -34,45 +47,82 @@ case class BackfillConfig(
...
   }
 
-  /**
-   * Get Milvus read options as a Map for DataSource
-   * Only reads the primary key field to minimize data transfer for join operation
-   */
+  /** Get Milvus read options as a Map for DataSource Only reads the primary key
+    * field to minimize data transfer for join operation
+    */
   def getMilvusReadOptions: Map[String, String] = {
```
liliu-z, line 126: getMilvusReadOptions unconditionally writes fs.access_key_id/value to FFI in IAM mode, and Properties.scala:51-64 still falls back to 'minioadmin'. Both layers must skip static credentials under IAM or client-mode Milvus FFI will fail with fake credentials.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -118,6 +177,7 @@ case class BackfillConfig(
       "fs.access_key_id" -> s3AccessKey,
```
liliu-z: Same issue as getMilvusReadOptions: AK/SK are unconditionally injected into FFI when writing segment paths. Skip static credential injection in IAM mode.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -191,51 +238,63 @@ object MilvusBackfill {
...
   private def readBackfillData(
       spark: SparkSession,
-      path: String
+      path: String,
+      config: BackfillConfig
   ): Either[BackfillError, DataFrame] = {
     try {
+      // Ensure Hadoop S3A is configured for the source bucket (may differ
+      // from the Milvus storage bucket) before reading the parquet.
+      configureHadoopS3ForPath(spark, path, config, isSource = true)
```
liliu-z, line 256: s3:// → s3a:// normalization is only applied in readSnapshotJson. readBackfillData here does not rewrite the scheme, causing the wrong FileSystem to be used under Hadoop 3.4.1.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -674,47 +822,178 @@ object MilvusBackfill {
...
+    * if you need a separate sink for results).
+    */
+  def writeResultJson(
+      spark: SparkSession,
+      result: BackfillResult,
+      outputPath: String,
+      config: BackfillConfig
+  ): Either[BackfillError, Unit] = {
     try {
+      configureHadoopS3ForPath(spark, outputPath, config, isSource = false)
```
liliu-z, line 879: writeResultJson does not normalize s3:// to s3a://, unlike readSnapshotJson. Normalization must cover snapshot, parquet, and output paths consistently.

src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -0,0 +1,128 @@
...
+        case Left(error) =>
+          System.err.println(s"Backfill FAILED: ${error.message}")
+          System.exit(1)
+      }
+    } finally {
+      spark.stop()
+    }
+  }
+
+  private[backfill] def parseArgs(args: Array[String]): Map[String, String] = {
```
liliu-z, line 135: parseArgs silently accepts any unknown --key, so typos like --s3access-key are absorbed and later ignored, causing confusing failures in the default provider chain. Validate keys against a whitelist.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -674,47 +822,178 @@ object MilvusBackfill {
...
+    * *source* bucket and the Milvus storage bucket (snapshot / segments) can
+    * each use their own endpoint and credentials within the same Spark session.
+    * No-op for non-S3 paths.
+    *
+    * @param isSource
+    *   true when configuring the backfill input bucket — in that case we
+    *   consult the `sourceS3*` overrides and fall back to the main credentials
+    *   when a particular field is unset.
+    */
+  private[backfill] def configureHadoopS3ForPath(
```
liliu-z, line 933: Hardcoded v1 provider class is unfriendly to IRSA/EKS. Consider IAMInstanceCredentialsProvider or WebIdentityTokenCredentialsProvider for K8s-native identity chains, and add a smoke test for this new path.

PR #80, opened 2026-04-14
enhance: adapt to new milvus-storage Transaction API with dropColumn for backfill column replacement

congqixia:
Bumps milvus-storage submodule to 6796b60, which refactors the Transaction JNI API: staging ops are now dedicated methods (appendFiles / addColumnGroups / dropColumn) and the old commit(updateId, resolveId, ptr) overload is gone.

Migrates MilvusLoonWriter.commit to the new API. The backfill (addfield) path now stages dropColumn for each target field before addColumnGroups, giving an atomic column-replace: first-time backfill is a noop drop plus an add (idempotent, matches prior behavior), while re-running backfill against the same manifest cleanly overwrites the previous column group instead of failing the schema-evolution validation. Normal writes take the appendFiles branch.

### DIFF
--- a/src/main/scala/write/MilvusLoonWriter.scala
+++ b/src/main/scala/write/MilvusLoonWriter.scala
@@ -281,23 +281,38 @@ class MilvusLoonPartitionWriter(
 
       // Commit column groups to manifest using Transaction
       val transaction = new MilvusStorageTransaction()
-      transaction.begin(basePath, writerProperties)
+      val committedVersion =
+        try {
+          transaction.begin(basePath, writerProperties)
 
-      milvusOption.options.get(
-        MilvusOption.WriterCommitType.toLowerCase
-      ) match {
-        case Some("addfield") =>
-          // Backfill = column replacement: drop each target column (noop if
-          // absent) then add the new column groups in the same transaction.
-          // Native commit orders DropColumn before AddColumnGroup validation,
-          // so this is atomic per-column overwrite.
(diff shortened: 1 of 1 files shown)


src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -283,14 +283,20 @@ class MilvusLoonPartitionWriter(
...
         MilvusOption.WriterCommitType.toLowerCase
       ) match {
-        case Some("addfield") => 1 // ADDFIELD
-        case _                => 0 // ADDFILES (default)
+        case Some("addfield") =>
+          // Backfill = column replacement: drop each target column (noop if
+          // absent) then add the new column groups in the same transaction.
+          // Native commit orders DropColumn before AddColumnGroup validation,
+          // so this is atomic per-column overwrite.
+          sparkSchema.fields.foreach(f => transaction.dropColumn(f.name))
```
liliu-z: transaction.destroy() is not protected by try-finally; any exception during dropColumn, addColumnGroups, appendFiles, or commit will leak native memory. This PR adds more failure points, making the leak more likely.

liliu-z: dropColumn iterates over all sparkSchema.fields, but it should only target backfill columns. This drops and re-adds every field in the schema, including ones that shouldn't be replaced (e.g. primary key), risking unnecessary rewrites or data integrity issues.

PR #81, opened 2026-04-14
enhance: support parquet column mapping for backfill

congqixia:
Add --column-mapping CLI flag (parquet_col:milvus_field,...) so backfill parquets no longer need a literal "pk" column and can carry extra or renamed columns. The mapping is the single source of truth: listed parquet columns are selected and renamed to their Milvus field names, unlisted columns are dropped, and the PK is auto-detected as the entry whose target matches the collection's primary key name.

Internally MilvusBackfill applies the mapping once (via a new applyColumnMapping helper) right after reading the parquet, which lets us drop the 4 hardcoded "pk" string references and switch performJoin to the using-column form. When --column-mapping is omitted, a legacy implicit mapping ({"pk" -> pkName} plus identity for other columns) is synthesized, so existing callers and scripts keep working unchanged.

### DIFF
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -143,7 +143,7 @@ object BackfillApp {
         "--column-mapping cannot be empty"
       )
     }
-    trimmed
+    val pairs = trimmed
       .split(",")
       .map(_.trim)
       .filter(_.nonEmpty)
@@ -157,7 +157,19 @@ object BackfillApp {
             )
         }
       }
-      .toMap
+      .toSeq
+    // Duplicate sources silently collapse under `.toMap`; fail fast, mirroring
+    // how applyColumnMapping rejects duplicate targets.
+    val dupSrc = pairs
+      .groupBy(_._1)
+      .collect { case (k, vs) if vs.size > 1 => k }
+      .toSeq
+    if (dupSrc.nonEmpty) {
+      throw new IllegalArgumentException(
+        s"--column-mapping has duplicate source columns: ${dupSrc.mkString(", ")}"
+      )
+    }
+    pairs.toMap
   }
 
   private[backfill] def parseArgs(args: Array[String]): Map[String, String] = {
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -67,7 +67,8 @@ object BackfillApp {
         if (parsed.contains("source-s3-use-ssl")) Some(true) else None,
       sourceS3UseIam = sourceUseIamFlag,
       sourceS3Region = parsed.get("source-s3-region"),
-      batchSize = parsed.getOrElse("batch-size", "1024").toInt
+      batchSize = parsed.getOrElse("batch-size", "1024").toInt,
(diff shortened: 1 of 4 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -291,6 +294,93 @@ object MilvusBackfill {
     }
   }
 
+  /** Project the raw backfill DataFrame through a parquet-column → Milvus-field
```
liliu-z, line 297: foldLeft + withColumnRenamed renames columns sequentially, so when a mapping target is also another mapping source (e.g. a→b, b→c) the second rename hits the already-renamed column, corrupting the schema. Use a single `select` with aliased columns instead.

src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -127,11 +128,38 @@ object BackfillApp {
...
+        entry.split(":", 2) match {
+          case Array(src, tgt) if src.nonEmpty && tgt.nonEmpty =>
+            src.trim -> tgt.trim
+          case _ =>
+            throw new IllegalArgumentException(
+              s"--column-mapping entry '$entry' must be of the form 'src:tgt'"
+            )
+        }
+      }
+      .toMap
```
liliu-z: parseColumnMapping calls `.toMap` which silently drops duplicate source keys. Duplicate sources should fail fast with an explicit error, same as duplicate targets already do.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -291,6 +294,93 @@ object MilvusBackfill {
...
+    * mapping so downstream code sees column names that match the Milvus schema
+    * exactly (including the PK, which must be named `pkName`).
+    *
+    * When `userMapping` is None, a legacy implicit mapping is synthesized: the
```
liliu-z, line 301: When `userMapping=None` and the Parquet already has a column matching `pkName`, the synthesized legacy mapping triggers a duplicate-target error referencing 'column mapping', which is confusing for users who never supplied `--column-mapping`. Add a dedicated check with a clearer message.

src/main/scala/operations/backfill/MilvusBackfill.scala
liliu-z, line 297: No tests exist for the new column mapping logic. parseColumnMapping and applyColumnMapping are scoped private[backfill] for testability but nothing exercises them—need coverage for normal mapping, legacy path, chain-rename conflicts, duplicate keys, missing pkName, and end-to-end join schema.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -291,6 +294,105 @@ object MilvusBackfill {
...
+              path = "",
+              message =
+                "Backfill parquet must contain a 'pk' column (or supply --column-mapping to rename the PK column)"
+            )
+          )
+        }
+        // If the parquet already has a column named pkName, the implicit
+        // {pk→pkName} rename would collide with it. Surface a dedicated error
+        // rather than letting the generic duplicate-target check fire and
+        // reference "column mapping" — users in the legacy path never passed
```
liliu-z, line 328: DataReadError(path="") discards the actual backfill data path, making diagnosis impossible. Should use SchemaValidationError or pass the real path.

PR #82, opened 2026-04-15
enhance: add --mode coalesce to backfill for fill-if-null semantics

congqixia:
Backfill previously always overwrote the target field with the user-provided parquet values (including nulls), which is correct for a fresh AddField but wrong as a repair tool: re-running backfill to plug null rows would also clobber non-null values written by an earlier run.

Add a --mode CLI parameter with two values:

    overwrite (default) — unchanged; parquet is the source of truth.
    coalesce            — read each target field from source and write the
                        parquet value only where the source value is null.
                        Per-field, per-row judgment.

Implementation:
- New constants BackfillMode{,Overwrite,Coalesce} in MilvusOption.
- BackfillConfig.mode with validator rejecting unknown values.
- BackfillApp wires --mode through KvFlags/parseArgs and updates usage.
- MilvusBackfill.run resolves newFieldNameToId earlier so readCollection WithMetadata can be asked to materialize target fields from source via an additional extraReadFields parameter (ReaderFieldIDs + snapshot schema extension).
- performJoin branches on mode: overwrite keeps the left-join; coalesce renames the backfill side with a __bf__ suffix, left-joins, then coalesce(src, bf) per field and drops the suffixed columns.
- Writer path is unchanged — the joined DF has the same shape in both modes.

A field with no prior binlogs reads as all-null and degrades naturally to overwrite semantics; no special branch.

Tests: 8 new cases in BackfillModeTest covering config/CLI validation, per-field independence (pk=1 keeps both src, pk=2 fills f1 keeps f2, pk=3 fills both, pk=4 no backfill row → src preserved), and the all-null degradation case. Full backfill unit suite (67 tests) passes.

### DIFF
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -66,6 +66,8 @@ object MilvusBackfill {
       case Right(_) => // Continue
     }
 
+    logger.info(s"Backfill mode: ${config.mode}")
+
     // Step 1: Try to load snapshot metadata
     val snapshotMetadataOpt =
       loadSnapshotMetadata(spark, snapshotPath, config) match {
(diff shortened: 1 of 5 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -140,12 +142,81 @@ object MilvusBackfill {
...
+              s"Available fields: ${fieldNameToId.keys.mkString(", ")}"
+          )
+        )
+      }
+      val newFieldNameToId = newFieldNames.map(n => n -> fieldNameToId(n)).toMap
+
+      // In coalesce mode, also read each target field from source so
+      // `coalesce(src, backfill)` can keep the existing value when non-null.
+      // Requires a snapshot — we need the field's Spark type.
+      val isCoalesceMode = config.mode == MilvusOption.BackfillModeCoalesce
```
liliu-z, line 180: Coalesce mode lacks per-field type validation between `MilvusSnapshotReader.fieldToSparkType` and the parquet schema. Mismatched types (e.g., IntegerType vs LongType) will be silently widened by `coalesce`, causing the writer to produce binlogs with incorrect types. Add an explicit type equality check for each field before calling `performJoin`.

PR #83, opened 2026-04-15
feat: [backfill]support StorageV2 (non-manifest packed parquet) snapshots

congqixia:
Backfill previously only handled the V3 (manifest-based) layout. Snapshots whose manifest_list carries milvus segcore's per-segment AVRO manifests — i.e. segment_info.storage_version == 2 — crashed during planning. This change makes the same MilvusBackfill.run() entry point work end-to-end on a V2 snapshot: read the PK column via the packed reader, join with the new-field input parquet, write one parquet per new field at the StorageV2 path convention, and emit a per-segment column_groups artifact in the result JSON ready for the caller to patch into the snapshot AVRO.

Why AVRO over the live REST/gRPC: GetPersistentSegmentInfo does not surface storage_version in this Milvus build, and REST /v2/vectordb/segments/describe exposes only the column-group slot, not real field ids. The snapshot AVRO carries storage_version authoritatively and is offline-friendly.

Read path:
- MilvusSegmentManifestReader: schemaless AVRO decoder using the verbatim schema from milvus/internal/datacoord/snapshot.go getProperAvroSchema(), bundled as src/main/resources/milvus-segment-manifest.avsc.
- MilvusParquetFooterReader: pulls storage_version + group_field_id_list from parquet kv-metadata.
- V2SegmentLoader: positional-joins AVRO binlog order with group_field_id_list to recover real (fieldIds, filePaths, fileRowCounts) per column group.
- MilvusLoonV2PartitionReader + milvus-storage's new MilvusStorageColumnGroups JNI: builds LoonColumnGroups* directly from (columns, files, fileRowCounts) without manifest resolution and feeds the packed reader. Uses logical column names ("ID", "RowID") to match what milvus segcore writes into V2 parquet.
- MilvusDataSource emits MilvusPackedV2InputPartitions driven by a new SnapshotV2Segments option populated by the backfill driver.

Write path:
- MilvusV2BinlogWriter: one AvroParquetWriter per new field, with kv-metadata storage_version="1.0.0" + group_field_id_list="{fid}" and Arrow PARQUET:field_id metadata. Path:
insert_log/{coll}/{part}/{seg}/{fid}/{logId} — the single-field column-group convention is slot = fieldId.
- MilvusBackfill.run() eagerly loads V2 segments once, threads them through readCollectionWithMetadata and extractMetadataFromSnapshot, and dispatches each segment to the V2 writer or the existing V3 Transaction+ADDFIELD path via a v2SegmentIdSet. Coexists with --mode coalesce: readCollectionWithMetadata also takes extraReadFields so the V2 read path materializes target fields from source when requested.

Artifact:
- BackfillResult.SegmentBackfillResult.v2Artifact carries V2SegmentArtifact; toJson emits storage_version + column_groups block matching the snapshot AVRO's ManifestEntry shape.

Dependencies:
- Add provided org.apache.avro:avro 1.12.0 (matches Spark 4.0's bundle).
- Bundle org.apache.parquet:parquet-avro 1.15.2 (Spark ships parquet-hadoop but not parquet-avro at this version).

milvus-storage (submodule bump):
- New column_groups JNI (createFromGroups / destroy, wrapped in extern "C") plus a Scala wrapper with resource-managed pointer.

Tooling:
- tools/ListV2SegmentsApp + examples/list_v2_segments.sh: standalone Spark job that prints decoded V2SegmentInfo per segment for read-side smoke testing.

Also:
- scalafmt reflow in MilvusClient, MilvusInsertDataWriter, and their tests (no behavior change).

Verification:
- 25/25 read-side unit tests pass; the existing V3 snapshot suite is untouched.
- backfill_demo_scala.sh runs end-to-end on a V2 snapshot in local minio and produces a result JSON with per-segment storage_version: 2 and a column_groups block.

Out of scope:
- Vector / Array[Float] new-field types in the V2 writer.
- Caller-injected logId allocator (currently a monotonic local seed).
- Read-back round-trip verification against the produced parquet


liliu-z:
**src/main/scala/zilliztech/spark/milvus/reader/MilvusLoonV2PartitionReader.scala:1** If the constructor fails after allocating columnGroupsPtr or arrowAllocator but before completing, Spark won't call close(), leaking native JNI and Arrow resources.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/reader/MilvusLoonV2PartitionReader.scala:1** close() releases columnGroupsPtr, loonReaderPtr, and arrowAllocator sequentially in a single try block; if an earlier release throws, the remaining resources leak. Each release needs its own independent try-catch.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/binlog/V2SegmentLoader.scala:86** head.binlogs.head called without checking if binlogs is empty; throws NoSuchElementException when a segment has no binlogs.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScanBuilder.scala:1** pushFilters() marks filters as supported on the V2 packed reader path, but that reader never applies them — Spark will skip filtering it expects was already done.

PR #84, opened 2026-04-16
fix: [backfill] bound V2 writer direct memory by releasing VectorSchemaRoot right after export

congqixia:
MilvusV2BinlogWriter and MilvusLoonWriter previously retained every flushed VectorSchemaRoot in `exportedRoots` until the native writer closed, because Arrow C Data Interface's zero-copy export shares buffer pointers with the C++ PackedRecordBatchWriter's cached RecordBatch shared_ptrs. Per-writer direct memory grew as O(totalRows × avgRowBytes), causing executors to OOM with `org.apache.arrow.memory.OutOfMemoryException: Failure allocating buffer` on large V2 segments.

Arrow 17's `Data.exportVectorSchemaRoot` retains each buffer independently of the source root, and C++'s `ImportRecordBatch` moves the release callback into its shared_ptr — so closing the source root right after export drops only the JVM-side ref while the buffer stays alive until C++ drops the cached batch. Direct memory is now bounded to roughly 16 MB × numGroups (the packed writer's buffer_size × column groups) regardless of segment size.

Verified via the new ArrowCDataRefcountTest: 12 round-trip cases covering Int32 / Int64 / Float32 / Bool / Utf8 / Binary / FixedSizeBinary / FixedSizeList<Float> / List<Int>, combined multi-type schema, sequential multi-batch, and concurrent-holders scenarios (all exported batches close their source roots immediately and are imported afterwards with intact data). All 12 tests pass on Arrow 17.0.0; allocator is empty after every import closes, proving no leak.

### DIFF
--- a/src/main/scala/write/MilvusLoonWriter.scala
+++ b/src/main/scala/write/MilvusLoonWriter.scala
@@ -217,7 +217,7 @@ class MilvusLoonPartitionWriter(
   private var totalRecordCount = 0L
 
   // Allocate initial capacity for vectors
-  allocateVectors()
+  allocateVectors(root)
 
   // Base path for writing - use custom path if provided, otherwise generate
   private val basePath = {
@@ -390,25 +390,39 @@ class MilvusLoonPartitionWriter(
       arrowArrayC.close()
     }
 
+    // Fully build + allocate the replacement root BEFORE swapping `root`, so
+    // that if allocation throws, the field still points at the old root and
+    // cleanup() can release it. Otherwise an allocation failure mid-swap would
+    // leak the half-allocated newRoot and forget the old one.
+    //
     // Old root's buffers are referenced only by C++ via the export's retained
-    // ref. Closing here drops the JVM ref — buffers stay alive until C++
-    // flushes its cached shared_ptr. Swap in a fresh root before closing so
-    // that if close() misbehaves, cleanup() still has a valid `root` handle.
+    // ref, so closing `oldRoot` at the end just drops the JVM ref — buffers
+    // stay alive until C++ flushes its cached shared_ptr.
+    val newRoot = VectorSchemaRoot.create(arrowSchema, allocator)
+    try {
+      allocateVectors(newRoot)
+    } catch {
+      case t: Throwable =>
+        try newRoot.close()
+        catch { case _: Throwable => /* swallow secondary cleanup error */ }
+        throw t
+    }
     val oldRoot = root
-    root = VectorSchemaRoot.create(arrowSchema, allocator)
-    allocateVectors()
+    root = newRoot
     currentBatchSize = 0
     oldRoot.close()
   }
 
-  /** Allocate or reallocate vectors for the current batch
+  /** Allocate or reallocate vectors for the given root. Takes the root as a
+    * parameter so callers can fully prepare a fresh VectorSchemaRoot before
+    * committing it to the `root` field.
     */
-  private def allocateVectors(): Unit = {
+  private def allocateVectors(r: VectorSchemaRoot): Unit = {
     import scala.collection.JavaConverters._
     import org.apache.arrow.vector.{VarCharVector, BaseVariableWidthVector}
 
     // For each vector, set appropriate initial capacity
-    root.getFieldVectors.asScala.foreach { vector =>
+    r.getFieldVectors.asScala.foreach { vector =>
       vector match {
         case varCharVector: VarCharVector =>
           // For VarChar vectors, allocate both row capacity and byte capacity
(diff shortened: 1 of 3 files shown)


src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -369,32 +373,32 @@ class MilvusLoonPartitionWriter(
...
+      // release is still here and firing it drops the export-side ref (source
+      // root still has one ref, so buffers remain alive for cleanup to free).
       arrowArrayC.close()
     }
+
+    // Old root's buffers are referenced only by C++ via the export's retained
+    // ref. Closing here drops the JVM ref — buffers stay alive until C++
+    // flushes its cached shared_ptr. Swap in a fresh root before closing so
+    // that if close() misbehaves, cleanup() still has a valid `root` handle.
+    val oldRoot = root
```
liliu-z, line 413: oldRoot leaks if allocateVectors throws after root is swapped to the new VSR, since cleanup() no longer tracks it. Fully allocate newRoot in a local before assigning root, then close oldRoot.

src/main/scala/write/MilvusV2BinlogWriter.scala
```diff
@@ -235,17 +238,24 @@ class MilvusV2BinlogWriter(
...
+      // holds one ref, so buffers remain alive until abort/cleanup runs).
       cArray.close()
     }
+
+    // Old root's buffers are now referenced only by C++ via the export's
+    // retained ref. Closing here just drops the JVM ref — buffers stay alive
+    // until C++ flushes the cached shared_ptr. Swap in a fresh root before
+    // closing so that if close() misbehaves, cleanup() still operates on a
+    // valid `root` handle.
+    val oldRoot = root
```
liliu-z, line 269: Same swap-before-close leak: if allocateVectors throws after root reassignment, oldRoot becomes untracked and leaks. Construct and allocate newRoot locally before assigning root.

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -196,17 +196,21 @@ class MilvusLoonPartitionWriter(
...
   // IMPORTANT: root must be var because we create a new one for each flush.
-  // The C++ writer caches RecordBatch shared_ptrs that reference Java-owned buffers
-  // via Arrow C Data Interface (zero-copy). Reusing the same root would overwrite
-  // buffer contents that C++ still references, causing data corruption.
+  // The C++ writer caches RecordBatch shared_ptrs that reference Java-owned
+  // buffers via Arrow C Data Interface (zero-copy). Reusing the same root
+  // would overwrite buffer contents that C++ still references, causing data
+  // corruption — so we swap in a fresh root per flush.
+  //
+  // After export, the source root is closed immediately. `exportVectorSchemaRoot`
```
liliu-z, line 206: Comment references ArrowCDataRefcountTest, but the test file is missing from this branch. Correctness relies on comments rather than tests covering Binary, FixedSizeBinary, List, Map, and nullable validity bitmaps.

src/main/scala/write/MilvusV2BinlogWriter.scala
```diff
@@ -235,17 +238,35 @@ class MilvusV2BinlogWriter(
...
+    val newRoot = VectorSchemaRoot.create(arrowSchema, allocator)
+    try {
+      allocateVectors(newRoot)
+    } catch {
+      case t: Throwable =>
+        try newRoot.close()
+        catch { case _: Throwable => /* swallow secondary cleanup error */ }
+        throw t
+    }
+    val oldRoot = root
```
liliu-z, line 269: Bare `oldRoot.close()` after successful native write/flush can throw and propagate out of flushBatch, causing Spark to retry the task and duplicate writes. Wrap in Try/recover with logError like cleanup() does (lines 523-540).

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -369,42 +373,56 @@ class MilvusLoonPartitionWriter(
...
+    val newRoot = VectorSchemaRoot.create(arrowSchema, allocator)
+    try {
+      allocateVectors(newRoot)
+    } catch {
+      case t: Throwable =>
+        try newRoot.close()
+        catch { case _: Throwable => /* swallow secondary cleanup error */ }
+        throw t
+    }
+    val oldRoot = root
```
liliu-z, line 413: Bare `oldRoot.close()` after successful native write can throw and escape flushBatch, triggering task retry and duplicate writes. Make it best-effort (Try/recover + logError) to match cleanup()'s strategy.

src/test/scala/write/ArrowCDataRefcountTest.scala
```diff
@@ -0,0 +1,720 @@
+package com.zilliz.spark.connector.write
```
liliu-z, line 1: Coverage claim overstates scope: real write path emits DoubleType, Short/Byte, Map, Struct, generic ArrayType, and FixedSizeBinary vectors — none tested here. Either add round-trips for Map/Struct/List<String>/Double or narrow the claim to primitive + fixed-width + single-level list.

PR #86, opened 2026-04-17
enhance: [backfill] report source / data-file / matched row counts in result and logs

congqixia:
Adds three levels of observability to MilvusBackfill so users can verify join completeness and diagnose low match rates:

- per-segment `sourceRowCount` and `matchedRowCount` on `SegmentBackfillResult`
- top-level `totalSourceRows`, `totalBackfillDataRows`, `totalMatchedRows` on `BackfillResult` (all surfaced in `summary`, `segmentSummary`, and `toJson`)
- driver-side INFO logs for totals + per-segment source/matched/rate

Matched rows are counted via a `__bf_matched__ = lit(true)` marker injected on the backfill side before the left join (same semantics in overwrite and coalesce modes: "source row whose PK was present in the data file"). The marker column is stripped from the writer projection in both V3 and V2 paths. Backfill file total uses a single `backfillDF.count()` up front.

### DIFF
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -36,6 +36,13 @@ object MilvusBackfill {
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  /** Marker column added to the backfill side before the left join so we can
+    * count how many source rows found a PK match (non-null marker → matched).
+    * Kept on the DataFrame all the way down to the per-segment partition
+    * function, and stripped from the projection written to parquet.
+    */
+  private[backfill] val MatchFlagCol = "__bf_matched__"
+
   /** Backfill new fields into a Milvus collection
     *
     * @param spark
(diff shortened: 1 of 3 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -156,6 +163,13 @@ object MilvusBackfill {
...
       }
 
+      // One extra parquet scan so the result can report the raw input row
+      // count alongside snapshot / matched counts — useful when debugging
+      // low join-match rates (e.g. data file has keys that aren't in the
+      // collection).
+      val backfillRowCount = backfillDF.count()
+      logger.info(s"Backfill data file rows: $backfillRowCount")
+
       // Extract new field names (all post-mapping columns except the PK).
```
liliu-z, line 176: No guard against user data containing the reserved column name `__bf_matched__`. In overwrite mode the user column is silently clobbered; in coalesce mode this causes an `AnalysisException` at runtime. Add a `require` check on `newFieldNames` to reject reserved names.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -156,6 +163,13 @@ object MilvusBackfill {
         case Right(df)   => df
       }
 
+      // One extra parquet scan so the result can report the raw input row
```
liliu-z: `backfillDF.count()` forces a full re-scan of uncached source data. Call `backfillDF.cache()` before `count()`, and move `unpersist()` to after `processSegments` completes (after L309), not after `performJoin()`.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -992,12 +1037,19 @@ object MilvusBackfill {
     try {
       var rowCount = 0L
       var nullRowCount = 0L
+      var matchedRowCount = 0L
 
       def writeRow(row: InternalRow): Unit = {
```
liliu-z, line 1096: `rowCount` from the join output is assigned to `sourceRowCount`, which is misleading when source PKs have duplicates — all derived metrics (match rate, completeness) become unreliable. Either fail fast on duplicate PKs or detect and flag them.

PR #87, opened 2026-04-19
fix: [v2backfill] correct read/write for backfilled segments and V2 column naming

congqixia:
Three independent correctness fixes along the V2 read/write path, all surfaced by backfill + reader integration:

1. V2SegmentLoader: per-entry field-id recovery. A segment that has been backfilled contains parquets from multiple write sessions; each parquet's footer-level `group_field_id_list` kv-metadata only advertises that session's groups. Reading one arbitrary footer and reusing its list for every AddField entry mis-attributes newly appended column groups. Recover field ids per entry from each parquet's own top-level schema instead — in StorageV2 one parquet == one column group, so the schema's columns ARE that group's field ids. Added `MilvusParquetFooterReader.readFieldIdsFromSchema` plus unit tests (normal multi-column, backfill-style single-column, and malformed no-field-id parquet).

2. MilvusV2BinlogWriter: keep logical column names in packed parquet. Milvus segcore's V2 packed parquet uses logical field names as parquet column names and carries the numeric fieldID only in `PARQUET:field_id` metadata. The V2 reader matches columns by logical name, so writing fieldID-as-string (V3 writer convention) produces files the reader cannot join. `SchemaUtil` gains a `useFieldIdAsName` flag (default true to preserve V3 behavior); the V2 writer sets it to false.

3. ArrowConverter: use `setSafe` for VarChar/VarBinary. `VarCharVector.set` / `VarBinaryVector.set` throw IndexOutOfBoundsException when the actual bytes exceed the density-based initial capacity (32 bytes per row). `setSafe` reallocates the data buffer instead — matches how the rest of the writer already behaves and removes a latent crash on longer strings/blobs.

### DIFF
--- a/src/main/scala/read/V2SegmentLoader.scala
+++ b/src/main/scala/read/V2SegmentLoader.scala
@@ -85,28 +85,34 @@ object V2SegmentLoader extends Logging {
             columnGroups = Seq.empty
           )
         } else {
-          // Any one of this segment's parquet files carries the full
-          // `group_field_id_list` in its kv-metadata. Skip any binlog_files
-          // entry that happens to be empty.
-          val sample = entry.binlogFiles
-            .find(_.binlogs.nonEmpty)
-            .get
-            .binlogs
-            .head
-            .logPath
-          val footerPath = resolvePath(sample, bucket)
-          val footer =
-            MilvusParquetFooterReader.read(footerPath, hadoopConf) match {
-              case Right(f) => f
-              case Left(err) =>
-                throw new RuntimeException(
-                  s"failed to read parquet footer $footerPath: ${err.getMessage}",
-                  err
-                )
+          // Per-entry field-id recovery: each V2 parquet file holds exactly
+          // one column group, and its schema's top-level columns ARE that
+          // group's field IDs. Reading per entry (rather than reusing a
+          // single footer's segment-level `group_field_id_list`) is required
+          // because a segment that has been backfilled contains parquets
+          // from multiple write sessions, each advertising only its own
+          // session's groups — see MilvusParquetFooterReader.readFieldIdsFromSchema.
+          val groupFieldIdListPerEntry: Seq[Seq[Long]] =
+            entry.binlogFiles.map { afb =>
+              if (afb.binlogs.isEmpty) Seq.empty
+              else {
+                val samplePath = resolvePath(afb.binlogs.head.logPath, bucket)
+                MilvusParquetFooterReader
+                  .readFieldIdsFromSchema(samplePath, hadoopConf) match {
+                  case Right(ids) => ids
+                  case Left(err) =>
+                    throw new RuntimeException(
+                      s"failed to read field ids from parquet $samplePath " +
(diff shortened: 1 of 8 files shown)


src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -85,28 +85,34 @@ object V2SegmentLoader extends Logging {
...
-                  err
-                )
+          // Per-entry field-id recovery: each V2 parquet file holds exactly
+          // one column group, and its schema's top-level columns ARE that
+          // group's field IDs. Reading per entry (rather than reusing a
+          // single footer's segment-level `group_field_id_list`) is required
+          // because a segment that has been backfilled contains parquets
+          // from multiple write sessions, each advertising only its own
+          // session's groups — see MilvusParquetFooterReader.readFieldIdsFromSchema.
+          val groupFieldIdListPerEntry: Seq[Seq[Long]] =
```
liliu-z: Mapping `afb.binlogs.isEmpty` to `Seq.empty` silently produces a V2ColumnGroup with empty fieldIds/filePaths/fileRowCounts. The top-level all-empty check doesn't guard against individual empty entries, so a corrupt manifest with partial-empty binlogs is now silently accepted instead of failing.

src/main/scala/serde/SchemaUtil.scala
```diff
@@ -228,13 +228,23 @@ object MilvusSchemaUtil {
...
+    *   while `PARQUET:field_id` metadata still carries the fieldID.
     * @return
     *   Arrow Schema
     */
   def convertSparkSchemaToArrow(
       sparkSchema: org.apache.spark.sql.types.StructType,
       vectorDimensions: Map[String, Int] = Map.empty,
-      fieldIds: Map[String, Long] = Map.empty
+      fieldIds: Map[String, Long] = Map.empty,
+      useFieldIdAsName: Boolean = true
```
liliu-z, line 247: No test coverage for `convertSparkSchemaToArrow(useFieldIdAsName = false)`. All existing tests use the default `true`; the new V2 path where logical column names are preserved is untested.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -85,28 +85,44 @@ object V2SegmentLoader extends Logging {
...
-                  s"failed to read parquet footer $footerPath: ${err.getMessage}",
-                  err
+          // Per-entry field-id recovery: each V2 parquet file holds exactly
+          // one column group, and its schema's top-level columns ARE that
+          // group's field IDs. Reading per entry (rather than reusing a
+          // single footer's segment-level `group_field_id_list`) is required
+          // because a segment that has been backfilled contains parquets
+          // from multiple write sessions, each advertising only its own
+          // session's groups — see MilvusParquetFooterReader.readFieldIdsFromSchema.
+          val groupFieldIdListPerEntry: Seq[Seq[Long]] =
```
liliu-z: The new per-entry schema recovery, partial-empty entry fail-fast, and enriched error context (segmentId/slotFieldId) paths have no direct unit tests; correctness depends solely on integration tests that require minio/S3. These branches should be covered by unit tests that don't require external infrastructure.

PR #89, opened 2026-04-21
fix: [reader] switch partition readers to per-batch RecordBatchReader to avoid Arrow Java offset bug

congqixia:
Arrow Java's `ArrowReader.loadNextBatch` (the C Data ArrowArrayStream path) imports every batch into the same VectorSchemaRoot and ignores `ArrowArray.offset`. When the underlying C++ `PackedRecordBatchReader` emits a sliced batch (the common case once a row group exceeds `min_rows` — the remainder is kept as `rb->Slice(min_rows)` in its per-chunk queue), every post-slice batch re-reads from buffer position 0. The observable symptom in the Spark connector was a row-cycle of period `min_rows`: e.g. primary keys repeated as `ID = row_offset mod 8192` while `row_offset` itself kept advancing, so a 20480-row segment surfaced only 8192 distinct IDs.

Pick up milvus-io/milvus-storage#493, which adds a per-batch FFI (`loon_record_batch_reader_new / read_next / destroy`) that exports each batch as a fresh offset-0 `ArrowArray`+`ArrowSchema` pair (columns with non-zero offsets are materialised via `arrow::Concatenate` on the C++ side), and rewire both partition readers onto it:

* Bump `milvus-storage` submodule 216e7ff → 84b433d.
* `MilvusLoonPartitionReader`: drop `ArrowArrayStream` / `arrowReader`; open an `rbrHandle` via `openRecordBatchReaderScala`, pull each batch with `readNextBatchScala` into a freshly-imported `VectorSchemaRoot`, close the previous root before importing the next. Same shape for the batch-mode scan path in `readAllRowsIntoInternalRows`.
* `MilvusLoonV2PartitionReader`: same migration; track a single live batch at a time to bound JVM direct memory, and release the `CDataDictionaryProvider` / handle in `releaseAll`.
* `MilvusStorageFFITest`: update the end-to-end test to drive the per-batch API.

Verified end-to-end against a V2 snapshot: a 5-group / 20480-row segment now reads monotonically-correct primary keys (`row_offset=8192 → ID=8192`, `row_offset=10000 → ID=10000`).

### DIFF
--- a/src/main/scala/read/MilvusLoonPartitionReader.scala
+++ b/src/main/scala/read/MilvusLoonPartitionReader.scala
@@ -32,6 +32,7 @@ import io.milvus.storage.{
   ArrowUtils,
   LatestColumnGroupsResult,
   MilvusStorageManifest,
+  MilvusStorageProperties,
   MilvusStorageReader,
   NativeLibraryLoader
 }
@@ -60,9 +61,6 @@ class MilvusLoonPartitionReader(
 
   private val sourceSchema = schema
 
-  // Create Arrow schema from Milvus schema
-  private val (arrowSchemaObj, arrowSchemaPtr) = createArrowSchema()
-
   private val fieldNameToId: Map[String, Long] = {
     val systemFields = Map("row_id" -> 0L, "timestamp" -> 1L)
     val userFields = milvusSchema.fields.map { field =>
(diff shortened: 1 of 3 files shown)


src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -114,26 +114,34 @@ class MilvusLoonPartitionReader(
...
-      (null, 0)
-    } else {
-      val batch = arrowReader.getVectorSchemaRoot
-      (batch, 0)
+  // Open the per-batch RecordBatchReader. Arrow Java's ArrowArrayStream path
+  // cannot be used here: its C Data importer ignores `ArrowArray.offset`, so
+  // sliced batches (the common case once a row group exceeds the reader's
+  // min_rows) would duplicate data. The per-batch API exports each batch as a
+  // fresh offset=0 ArrowArray+ArrowSchema pair.
+  private val rbrHandle: Long = reader.openRecordBatchReaderScala(null)
```
liliu-z: rbrHandle is a val, so it can't be nulled after close(). This makes close() non-idempotent and risks a double-free of the native handle.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -114,26 +114,34 @@ class MilvusLoonPartitionReader(
...
+      sch.close()
     }
   }
 
-  // Need mutable versions for iteration
-  private var _currentBatch: VectorSchemaRoot = currentBatch
-  private var _currentRowIndex: Int = currentRowIndex
+  // Eagerly load the first batch so empty segments short-circuit upfront.
+  private var _currentBatch: VectorSchemaRoot = pullNextBatch()
+  private var _currentRowIndex: Int = 0
```
liliu-z: close() uses a single try block: if any one step (e.g. rbrHandle.close()) throws, all subsequent resource releases are skipped, causing cascade leaks. Wrap each resource release in its own try-catch, as V2 already does.

src/main/scala/read/MilvusLoonV2PartitionReader.scala
```diff
@@ -142,31 +153,70 @@ class MilvusLoonV2PartitionReader(
...
 
-  private var _currentBatch: VectorSchemaRoot = {
-    if (arrowReader.loadNextBatch()) arrowReader.getVectorSchemaRoot else null
-  }
+  // Per-batch state: each loadNextBatch() imports a FRESH RecordBatch
+  // into a NEW VectorSchemaRoot. The previous root (if any) is closed
+  // before the next import so buffers are correctly reclaimed. This
+  // intentionally does not reuse a single root across batches — Arrow
+  // Java's shared-root import drops the per-batch `offset`.
+  private var _currentBatch: VectorSchemaRoot = loadNextBatch()
```
liliu-z: _currentBatch = loadNextBatch() is outside the constructor's try-catch (lines 111-158). If it throws, the constructor never completes so Spark won't call close(), leaking rbrHandle, dictProvider, reader, columnGroupsPtr, arrowSchemaObj, and readerProperties. Move this inside the existing try block or add a catch that calls releaseAll().

src/main/scala/read/MilvusLoonPartitionReader.scala
liliu-z: V1 constructor has no try-catch around initialization. If any allocation fails mid-construction, previously allocated native handles leak. This PR adds rbrHandle as another native resource at risk, making this worse.

src/test/scala/loon/MilvusStorageFFITest.scala
```diff
@@ -110,35 +110,55 @@ class MilvusStorageFFITest extends AnyFunSuite with Matchers {
         )
         info("✓ Reader created\n")
 
-        // Read data using Arrow C Data Interface
-        val recordBatchReaderPtr = reader.getRecordBatchReaderScala()
-        val arrowArrayStream = ArrowArrayStream.wrap(recordBatchReaderPtr)
-
+        // Read data via the per-batch RecordBatchReader (the only Java-safe
```
liliu-z, line 113: numRows=100 never crosses the sliced batch boundary, so the old buggy implementation would also pass. This test needs to explicitly cross the min_rows boundary and verify correctness at those points to serve as a regression test for the Arrow offset fix.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -175,18 +214,19 @@ class MilvusLoonPartitionReader(
...
+          if (_currentBatch != null) {
+            _currentBatch.close()
+            _currentBatch = null
+          }
+          _currentBatch = pullNextBatch()
+          _currentRowIndex = 0
+          if (_currentBatch == null) {
             // No more batches
             return false
+          } else if (_currentBatch.getRowCount == 0) {
```
liliu-z: V1 `next()` returns false on an empty batch without looping, so an empty batch is treated as EOF. If the native layer ever yields an empty batch before non-empty ones, all subsequent data is silently dropped. Adopt V2's while-loop pattern to skip empty batches.

PR #91, opened 2026-04-21
enhance!: [backfill] default --mode to coalesce for safer fill-if-null semantics

congqixia:
Flip BackfillConfig.mode and BackfillApp's --mode fallback from
"overwrite" to "coalesce". Coalesce only writes the parquet value when
the existing source field is null, so a re-run or repair job no longer
silently overwrites rows that already have data — the previous default
made accidental wipes too easy.

When --mode is omitted, BackfillApp now emits a startup notice on
stderr pointing at the change and the legacy opt-out, so operators spot the default flip in driver logs even if they don't read the release notes.

BREAKING CHANGE: the default value of --mode / BackfillConfig.mode has changed from "overwrite" to "coalesce".

Migration guidance for downstream callers that relied on the old default:

- Add --mode overwrite (CLI) or mode = MilvusOption.BackfillModeOverwrite (programmatic API) to restore the pre-https://github.com/zilliztech/spark-milvus/pull/91 "parquet wins" semantics.
- Client-mode callers (no --snapshot) MUST opt in to overwrite: coalesce needs the snapshot to read each target field and will otherwise fail validation with "--mode=coalesce requires a  snapshot path".
- Coalesce also enforces exact parquet ↔ collection field type match  (no Spark widening such as Int32 → Int64). Fix the parquet schema or switch back to overwrite.

Docs and the backfill README are updated with a Merge modes section, an Optional flags table that names --mode, and an updated programmatic-API example. The existing "BackfillConfig defaults to overwrite mode" unit test is renamed and now asserts the coalesce default; all other mode tests already set mode explicitly and remain
unchanged.

### DIFF
--- a/src/main/scala/operations/backfill/BackfillConfig.scala
+++ b/src/main/scala/operations/backfill/BackfillConfig.scala
@@ -79,12 +79,12 @@ case class BackfillConfig(
     columnMapping: Option[Map[String, String]] = None,
 
     // Merge mode for backfill values:
-    //   "overwrite" (default) — parquet is the source of truth; for every row
-    //     in the collection, write the parquet value (null included).
-    //   "coalesce" — read the target field's current value from source and
-    //     only write the parquet value where the source value is null.
-    //     Per-field: each target field is decided independently.
-    mode: String = MilvusOption.BackfillModeOverwrite
+    //   "coalesce" (default) — read the target field's current value from
+    //     source and only write the parquet value where the source value is
+    //     null. Per-field: each target field is decided independently.
+    //   "overwrite" — parquet is the source of truth; for every row in the
+    //     collection, write the parquet value (null included).
+    mode: String = MilvusOption.BackfillModeCoalesce
 ) {
 
   /** Validate S3 and writer configuration (always required)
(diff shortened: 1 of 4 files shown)


liliu-z:
**src/main/scala/operations/backfill/README.md:69** The `--mode` parameter is completely undocumented—not in CLI examples, not in the flags table, and not in the configuration reference. This is critical now that the default is changing and affects all users.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -79,12 +79,12 @@ case class BackfillConfig(
...
-    mode: String = MilvusOption.BackfillModeOverwrite
+    //   "coalesce" (default) — read the target field's current value from
+    //     source and only write the parquet value where the source value is
+    //     null. Per-field: each target field is decided independently.
+    //   "overwrite" — parquet is the source of truth; for every row in the
+    //     collection, write the parquet value (null included).
+    mode: String = MilvusOption.BackfillModeCoalesce
 ) {
 
   /** Validate S3 and writer configuration (always required)
```
liliu-z, line 90: Changing the default from `overwrite` to `coalesce` is a silent breaking change: existing callers that omit `--mode` will get fill-null-only semantics instead of full-overwrite (or fail outright without a snapshot). This needs to be called out as a breaking change with migration guidance to add `--mode overwrite` explicitly.

PR #94, opened 2026-05-14
enhance: [reader] add snapshot-backed read entrypoint

congqixia:
Add MilvusReadApp and a spark-submit wrapper for client and snapshot reads, and prefer client-created snapshots with fallback only when snapshot RPCs are unavailable.

Reuse snapshot planning for V2/V3 segments, expose real row_id/timestamp plus $segment_id/$row_offset metadata, and make Arrow string conversion handle JSON stored as binary.

Update Milvus proto for snapshot RPCs and add focused tests for the read app, snapshot planner helpers, and Arrow VarBinary StringType conversion.

Also: reader/backfill metadata columns are now named $segment_id and $row_offset instead of segment_id and row_offset.

### DIFF
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -59,21 +59,20 @@ import io.milvus.grpc.schema.{
   ValueField
 }
 
-import io.grpc._
-import io.grpc.{ClientInterceptor, Metadata, Status => GrpcStatus}
+import io.grpc.{
+  ClientInterceptor,
+  ManagedChannel,
+  Metadata,
+  Status => GrpcStatus,
+  StatusException,
+  StatusRuntimeException
+}
 import io.grpc.netty.shaded.io.grpc.netty.{GrpcSslContexts, NettyChannelBuilder}
 import io.grpc.stub.MetadataUtils
-import io.grpc.Status.Code
 
 /** A simplified client for interacting with Milvus
   */
 class MilvusClient(params: MilvusConnectionParams) extends Logging {
-  private val retryInterceptor = new GrpcRetryInterceptor(
-    maxRetries = 5,
-    initialDelayMillis = 500,
-    delayMultiplier = 2.0,
-    maxDelayMillis = 5000
-  )
   private lazy val channel: ManagedChannel = {
     val uri = new URI(params.uri)
     val scheme = uri.getScheme
@@ -88,10 +87,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       }
     }
 
-    val interceptors = Seq(
-      getConnectionMetadataInterceptor(),
-      retryInterceptor
-    )
+    val interceptors = Seq(getConnectionMetadataInterceptor())
     var channelBuilder = if (params.serverPemPath.nonEmpty) {
       val sslContext = GrpcSslContexts
         .forClient()
@@ -940,31 +936,42 @@ object MilvusClient {
   val RateLimitReasonMarker: String = "rate limit exceeded"
   val ServiceNotImplementedMarker: String = "service not implemented"
 
+  private val SnapshotRpcNames: Set[String] = Set(
+    "createsnapshot",
+    "describesnapshot",
+    "dropsnapshot"
+  )
+
+  private val ServiceUnavailableMarkers: Set[String] = Set(
+    ServiceNotImplementedMarker,
+    "unknown method",
+    "method not registered"
+  )
+
+  private def grpcStatus(t: Throwable): Option[GrpcStatus] = t match {
+    case e: StatusRuntimeException => Some(e.getStatus)
+    case e: StatusException        => Some(e.getStatus)
+    case _                         => None
+  }
+
+  private def isSnapshotRpcUnavailable(description: String): Boolean = {
+    val normalized = description.toLowerCase
+    SnapshotRpcNames.exists(normalized.contains) &&
+    ServiceUnavailableMarkers.exists(normalized.contains)
+  }
+
   def isServiceNotImplemented(t: Throwable): Boolean = {
     val visited = java.util.Collections.newSetFromMap(
       new java.util.IdentityHashMap[Throwable, java.lang.Boolean]()
     )
     var current = t
     while (current != null && visited.add(current)) {
-      current match {
-        case e: StatusRuntimeException
-            if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
-          return true
-        case e: StatusException
-            if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
-          return true
-        case _ =>
-      }
-
-      val message = Option(current.getMessage).getOrElse("").toLowerCase
-      val isGrpcException = current.isInstanceOf[StatusRuntimeException] ||
-        current.isInstanceOf[StatusException]
-      if (
-        isGrpcException &&
-        (message.contains(ServiceNotImplementedMarker) ||
-          message.contains("unknown method"))
-      ) {
-        return true
+      grpcStatus(current).foreach { status =>
+        if (status.getCode == GrpcStatus.Code.UNIMPLEMENTED) return true
+        if (status.getCode == GrpcStatus.Code.UNKNOWN) {
+          val description = Option(status.getDescription).getOrElse("")
+          if (isSnapshotRpcUnavailable(description)) return true
+        }
       }
       current = current.getCause
     }
@@ -1076,89 +1083,5 @@ object PKProcessor {
 
   implicit object StringProcessor extends PKProcessor[String] {
     def process(seq: Seq[String]): String = seq.map(s => s"'$s'").mkString(", ")
-  }
-}
-
-class GrpcRetryInterceptor(
-    maxRetries: Int = 5,
-    initialDelayMillis: Long = 500,
-    delayMultiplier: Double = 2.0,
-    maxDelayMillis: Long = 5000
-) extends ClientInterceptor {
-
-  private val nonRetryableCodes: Set[Code] = Set(
-    Code.DEADLINE_EXCEEDED,
-    Code.PERMISSION_DENIED,
-    Code.UNAUTHENTICATED,
-    Code.INVALID_ARGUMENT,
-    Code.ALREADY_EXISTS,
-    Code.RESOURCE_EXHAUSTED,
-    Code.UNIMPLEMENTED
-  )
-
-  override def interceptCall[ReqT, RespT](
-      method: MethodDescriptor[ReqT, RespT],
-      callOptions: CallOptions,
-      next: Channel
-  ): ClientCall[ReqT, RespT] = {
-    new ForwardingClientCall.SimpleForwardingClientCall[ReqT, RespT](
-      next.newCall(method, callOptions)
-    ) {
-      override def start(
-          responseListener: ClientCall.Listener[RespT],
-          headers: Metadata
-      ): Unit = {
-        var currentAttempt = 0
-        var currentDelay = initialDelayMillis
-
-        def executeCall(): Unit = {
-          currentAttempt += 1
-          println(
-            s"Attempting gRPC call for method ${method.getFullMethodName()}, attempt $currentAttempt"
-          )
-
-          val originalListener =
-            new ForwardingClientCallListener.SimpleForwardingClientCallListener[
-              RespT
-            ](responseListener) {
-              override def onClose(
-                  status: GrpcStatus,
-                  trailers: Metadata
-              ): Unit = {
-                if (status.isOk) {
-                  // Call succeeded
-                  super.onClose(status, trailers)
-                } else {
-                  val statusCode = status.getCode
-                  if (nonRetryableCodes.contains(statusCode)) {
-                    println(
-                      s"gRPC call failed with non-retryable status: $statusCode. Not retrying."
-                    )
-                    super.onClose(status, trailers)
-                  } else if (currentAttempt < maxRetries) {
-                    println(
-                      s"gRPC call failed with retryable status: $statusCode. Retrying in $currentDelay ms."
-                    )
-                    Thread.sleep(currentDelay)
-                    currentDelay = Math.min(
-                      (currentDelay * delayMultiplier).toLong,
-                      maxDelayMillis
-                    )
-                    executeCall()
-                  } else {
-                    println(
-                      s"gRPC call failed after $maxRetries attempts with status: $statusCode. No more retries."
-                    )
-                    super.onClose(status, trailers)
-                  }
-                }
-              }
-            }
-          super.start(originalListener, headers)
-        }
-
-        executeCall() // Start the first attempt
-      }
-    }
   }
 }
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -920,8 +920,9 @@ object MilvusClient {
       val isGrpcException = current.isInstanceOf[StatusRuntimeException] ||
         current.isInstanceOf[StatusException]
       if (
-        message.contains(ServiceNotImplementedMarker) ||
-        (isGrpcException && message.contains("unknown method"))
+        isGrpcException &&
+        (message.contains(ServiceNotImplementedMarker) ||
+          message.contains("unknown method"))
       ) {
         return true
       }
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -139,22 +139,26 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
     val server = MilvusServiceGrpc
       .blockingStub(channel)
       .withWaitForReady()
+    server
       .withDeadlineAfter(10, TimeUnit.SECONDS)
-    server.connect(
-      ConnectRequest(
-        clientInfo = Some(
-          ClientInfo(
-            sdkType = "spark-connector",
-            sdkVersion = "0.1.0",
-            localTime = java.time.LocalDateTime.now().toString,
-            host = java.net.InetAddress.getLocalHost.getHostName,
-            user = "scala-sdk-user"
+      .connect(
+        ConnectRequest(
+          clientInfo = Some(
+            ClientInfo(
+              sdkType = "spark-connector",
+              sdkVersion = "0.1.0",
+              localTime = java.time.LocalDateTime.now().toString,
+              host = java.net.InetAddress.getLocalHost.getHostName,
+              user = "scala-sdk-user"
+            )
           )
         )
       )
-    )
     server
   }
+
+  private def rpcStub: MilvusServiceGrpc.MilvusServiceBlockingStub =
+    stub.withDeadlineAfter(10, TimeUnit.SECONDS)
 
   private lazy val httpClient: HttpClient = {
     HttpClient
@@ -202,7 +206,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       properties: Map[String, String] = Map.empty
   ): Try[Status] = {
     try {
-      val status = stub.createDatabase(
+      val status = rpcStub.createDatabase(
         CreateDatabaseRequest(
           dbName = dbName,
           properties = properties
@@ -287,7 +291,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       properties: Map[String, String] = Map.empty
   ): Try[Status] = {
     try {
-      val status = stub.createCollection(
+      val status = rpcStub.createCollection(
         CreateCollectionRequest(
           dbName = dbName,
           collectionName = collectionName,
@@ -314,7 +318,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       collectionName: String
   ): Try[Status] = {
     try {
-      val status = stub.dropCollection(
+      val status = rpcStub.dropCollection(
         DropCollectionRequest(
           dbName = dbName,
           collectionName = collectionName
@@ -334,7 +338,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       collectionNames: Seq[String] = Seq.empty
   ): Try[Status] = {
     try {
-      val flushResponse = stub.flush(
+      val flushResponse = rpcStub.flush(
         FlushRequest(
           dbName = dbName,
           collectionNames = collectionNames
@@ -375,7 +379,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       schemaTimestamp: Long = 0L
   ): Try[Status] = {
     try {
-      val insertResult = stub.insert(
+      val insertResult = rpcStub.insert(
         InsertRequest(
           dbName = dbName,
           collectionName = collectionName,
@@ -411,7 +415,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       rowBased: Boolean = false
   ): Try[Seq[Long]] = {
     try {
-      val importResult = stub.`import`(
+      val importResult = rpcStub.`import`(
         ImportRequest(
           dbName = dbName,
           collectionName = collectionName,
@@ -448,7 +452,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
 
   def getImportState(taskId: Long): Try[GetImportStateResponse] = {
     try {
-      val importStateResult = stub.getImportState(
+      val importStateResult = rpcStub.getImportState(
         GetImportStateRequest(task = taskId)
       )
       val status = importStateResult.status.getOrElse(
@@ -497,7 +501,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
           s"$name in [${processor.process(pks)}]"
         }
       }
-      val deleteResult = stub.delete(
+      val deleteResult = rpcStub.delete(
         DeleteRequest(
           dbName = dbName,
           collectionName = collectionName,
@@ -526,7 +530,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       dbName: String,
       collectionName: String
   ): DescribeCollectionResponse = {
-    return stub.describeCollection(
+    return rpcStub.describeCollection(
       DescribeCollectionRequest(
         dbName = dbName,
         collectionName = collectionName
@@ -630,7 +634,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       compactionProtectionSeconds: Long
   ): Try[MilvusSnapshotInfo] = {
     try {
-      val createStatus = stub.createSnapshot(
+      val createStatus = rpcStub.createSnapshot(
         CreateSnapshotRequest(
           name = snapshotName,
           description = description,
@@ -685,7 +689,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
     var lastFailure = Option.empty[Exception]
     (1 to maxAttempts).foreach { attempt =>
       try {
-        val snapshot = stub.describeSnapshot(
+        val snapshot = rpcStub.describeSnapshot(
           DescribeSnapshotRequest(
             name = snapshotName,
             dbName = dbName,
@@ -725,7 +729,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       snapshotName: String
   ): Try[Unit] = {
     try {
-      val status = stub.dropSnapshot(
+      val status = rpcStub.dropSnapshot(
         DropSnapshotRequest(
           name = snapshotName,
           dbName = dbName,
@@ -744,7 +748,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       collectionName: String
   ): Try[Seq[MilvusSegmentInfo]] = {
     try {
-      val segments = stub.getPersistentSegmentInfo(
+      val segments = rpcStub.getPersistentSegmentInfo(
         GetPersistentSegmentInfoRequest(
           dbName = dbName,
           collectionName = collectionName
@@ -869,7 +873,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       partitionName: String
   ): Try[Long] = {
     try {
-      val partitionInfos = stub.showPartitions(
+      val partitionInfos = rpcStub.showPartitions(
         ShowPartitionsRequest(
           dbName = dbName,
           collectionName = collectionName
@@ -898,7 +902,7 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       collectionName: String
   ): Try[Seq[MilvusPartitionInfo]] = {
     try {
-      val partitionInfos = stub.showPartitions(
+      val partitionInfos = rpcStub.showPartitions(
         ShowPartitionsRequest(
           dbName = dbName,
           collectionName = collectionName
@@ -1140,7 +1144,6 @@ class GrpcRetryInterceptor(
                       (currentDelay * delayMultiplier).toLong,
                       maxDelayMillis
                     )
-                    super.onClose(status, trailers)
                     executeCall()
                   } else {
                     println(
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -37,6 +37,7 @@ import io.milvus.grpc.milvus.{
   DescribeCollectionResponse,
   DescribeSnapshotRequest,
   DropCollectionRequest,
+  DropSnapshotRequest,
   FlushRequest,
   GetImportStateRequest,
   GetImportStateResponse,
@@ -638,33 +639,59 @@ class MilvusClient(params: MilvusConnectionParams) {
       )
       checkStatus("createSnapshot", createStatus).get
 
-      val snapshot = stub.describeSnapshot(
-        DescribeSnapshotRequest(
+      try {
+        val snapshot = stub.describeSnapshot(
+          DescribeSnapshotRequest(
+            name = snapshotName,
+            dbName = dbName,
+            collectionName = collectionName
+          )
+        )
+        checkStatus(
+          "describeSnapshot",
+          snapshot.status.getOrElse(
+            Status(
+              errorCode = ErrorCode.UnexpectedError,
+              reason = "DescribeSnapshot Status is empty"
+            )
+          )
+        ).get
+
+        Success(
+          MilvusSnapshotInfo(
+            name = snapshot.name,
+            description = snapshot.description,
+            collectionName = snapshot.collectionName,
+            partitionNames = snapshot.partitionNames,
+            createTs = snapshot.createTs,
+            s3Location = snapshot.s3Location
+          )
+        )
+      } catch {
+        case e: Exception =>
+          dropSnapshot(dbName, collectionName, snapshotName)
+          throw e
+      }
+    } catch {
+      case e: StatusRuntimeException => Failure(e)
+      case e: Exception              => Failure(e)
+    }
+  }
+
+  def dropSnapshot(
+      dbName: String,
+      collectionName: String,
+      snapshotName: String
+  ): Try[Unit] = {
+    try {
+      val status = stub.dropSnapshot(
+        DropSnapshotRequest(
           name = snapshotName,
           dbName = dbName,
           collectionName = collectionName
         )
       )
-      checkStatus(
-        "describeSnapshot",
-        snapshot.status.getOrElse(
-          Status(
-            errorCode = ErrorCode.UnexpectedError,
-            reason = "DescribeSnapshot Status is empty"
-          )
-        )
-      ).get
-
-      Success(
-        MilvusSnapshotInfo(
-          name = snapshot.name,
-          description = snapshot.description,
-          collectionName = snapshot.collectionName,
-          partitionNames = snapshot.partitionNames,
-          createTs = snapshot.createTs,
-          s3Location = snapshot.s3Location
-        )
-      )
+      checkStatus("dropSnapshot", status).map(_ => ())
     } catch {
       case e: StatusRuntimeException => Failure(e)
       case e: Exception              => Failure(e)
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -18,6 +18,7 @@ import com.fasterxml.jackson.module.scala.{
   ScalaObjectMapper
 }
 import com.google.protobuf.ByteString
+import org.apache.spark.internal.Logging
 
 import io.milvus.grpc.common.{
   ClientInfo,
@@ -62,7 +63,6 @@ import io.grpc.{ClientInterceptor, Metadata, Status => GrpcStatus}
 import io.grpc.netty.shaded.io.grpc.netty.{GrpcSslContexts, NettyChannelBuilder}
 import io.grpc.stub.MetadataUtils
 import io.grpc.Status.Code
-import org.apache.spark.internal.Logging
 
 /** A simplified client for interacting with Milvus
   */
@@ -904,8 +904,11 @@ object MilvusClient {
   val ServiceNotImplementedMarker: String = "service not implemented"
 
   def isServiceNotImplemented(t: Throwable): Boolean = {
+    val visited = java.util.Collections.newSetFromMap(
+      new java.util.IdentityHashMap[Throwable, java.lang.Boolean]()
+    )
     var current = t
-    while (current != null) {
+    while (current != null && visited.add(current)) {
       current match {
         case e: StatusRuntimeException
             if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -62,10 +62,11 @@ import io.grpc.{ClientInterceptor, Metadata, Status => GrpcStatus}
 import io.grpc.netty.shaded.io.grpc.netty.{GrpcSslContexts, NettyChannelBuilder}
 import io.grpc.stub.MetadataUtils
 import io.grpc.Status.Code
+import org.apache.spark.internal.Logging
 
 /** A simplified client for interacting with Milvus
   */
-class MilvusClient(params: MilvusConnectionParams) {
+class MilvusClient(params: MilvusConnectionParams) extends Logging {
   private val retryInterceptor = new GrpcRetryInterceptor(
     maxRetries = 5,
     initialDelayMillis = 500,
@@ -669,7 +670,14 @@ class MilvusClient(params: MilvusConnectionParams) {
         )
       } catch {
         case e: Exception =>
-          dropSnapshot(dbName, collectionName, snapshotName)
+          dropSnapshot(dbName, collectionName, snapshotName) match {
+            case Failure(dropErr) =>
+              logWarning(
+                s"Failed to drop snapshot $snapshotName after describeSnapshot failure",
+                dropErr
+              )
+            case Success(_) =>
+          }
           throw e
       }
     } catch {
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -884,7 +884,6 @@ object MilvusClient {
       val message = Option(current.getMessage).getOrElse("").toLowerCase
       if (
         message.contains(ServiceNotImplementedMarker) ||
-        message.contains("unimplemented") ||
         message.contains("unknown method")
       ) {
         return true
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -37,6 +37,7 @@ import io.milvus.grpc.milvus.{
   DescribeCollectionRequest,
   DescribeCollectionResponse,
   DescribeSnapshotRequest,
+  DescribeSnapshotResponse,
   DropCollectionRequest,
   DropSnapshotRequest,
   FlushRequest,
@@ -641,22 +642,11 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       checkStatus("createSnapshot", createStatus).get
 
       try {
-        val snapshot = stub.describeSnapshot(
-          DescribeSnapshotRequest(
-            name = snapshotName,
-            dbName = dbName,
-            collectionName = collectionName
-          )
+        val snapshot = describeSnapshotForReadWithRetry(
+          dbName,
+          collectionName,
+          snapshotName
         )
-        checkStatus(
-          "describeSnapshot",
-          snapshot.status.getOrElse(
-            Status(
-              errorCode = ErrorCode.UnexpectedError,
-              reason = "DescribeSnapshot Status is empty"
-            )
-          )
-        ).get
 
         Success(
           MilvusSnapshotInfo(
@@ -684,6 +674,49 @@ class MilvusClient(params: MilvusConnectionParams) extends Logging {
       case e: StatusRuntimeException => Failure(e)
       case e: Exception              => Failure(e)
     }
+  }
+
+  private def describeSnapshotForReadWithRetry(
+      dbName: String,
+      collectionName: String,
+      snapshotName: String,
+      maxAttempts: Int = 3
+  ): DescribeSnapshotResponse = {
+    var lastFailure = Option.empty[Exception]
+    (1 to maxAttempts).foreach { attempt =>
+      try {
+        val snapshot = stub.describeSnapshot(
+          DescribeSnapshotRequest(
+            name = snapshotName,
+            dbName = dbName,
+            collectionName = collectionName
+          )
+        )
+        checkStatus(
+          "describeSnapshot",
+          snapshot.status.getOrElse(
+            Status(
+              errorCode = ErrorCode.UnexpectedError,
+              reason = "DescribeSnapshot Status is empty"
+            )
+          )
+        ).get
+        return snapshot
+      } catch {
+        case e: Exception =>
+          lastFailure = Some(e)
+          if (attempt < maxAttempts) {
+            logWarning(
+              s"describeSnapshot failed for $snapshotName (attempt $attempt/$maxAttempts)",
+              e
+            )
+            TimeUnit.MILLISECONDS.sleep(200L * attempt)
+          }
+      }
+    }
+    throw lastFailure.getOrElse(
+      new RuntimeException(s"describeSnapshot failed for $snapshotName")
+    )
   }
 
   def dropSnapshot(
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -882,9 +882,11 @@ object MilvusClient {
       }
 
       val message = Option(current.getMessage).getOrElse("").toLowerCase
+      val isGrpcException = current.isInstanceOf[StatusRuntimeException] ||
+        current.isInstanceOf[StatusException]
       if (
         message.contains(ServiceNotImplementedMarker) ||
-        message.contains("unknown method")
+        (isGrpcException && message.contains("unknown method"))
       ) {
         return true
       }
--- a/src/main/scala/MilvusClient.scala
+++ b/src/main/scala/MilvusClient.scala
@@ -31,9 +31,11 @@ import io.milvus.grpc.milvus.{
   ConnectRequest,
   CreateCollectionRequest,
   CreateDatabaseRequest,
+  CreateSnapshotRequest,
   DeleteRequest,
   DescribeCollectionRequest,
   DescribeCollectionResponse,
+  DescribeSnapshotRequest,
   DropCollectionRequest,
   FlushRequest,
   GetImportStateRequest,
@@ -617,6 +619,58 @@ class MilvusClient(params: MilvusConnectionParams) {
     }
   }
 
+  def createSnapshotForRead(
+      dbName: String,
+      collectionName: String,
+      snapshotName: String,
+      description: String,
+      compactionProtectionSeconds: Long
+  ): Try[MilvusSnapshotInfo] = {
+    try {
+      val createStatus = stub.createSnapshot(
+        CreateSnapshotRequest(
+          name = snapshotName,
+          description = description,
+          dbName = dbName,
+          collectionName = collectionName,
+          compactionProtectionSeconds = compactionProtectionSeconds
+        )
+      )
+      checkStatus("createSnapshot", createStatus).get
+
+      val snapshot = stub.describeSnapshot(
+        DescribeSnapshotRequest(
+          name = snapshotName,
+          dbName = dbName,
+          collectionName = collectionName
+        )
+      )
+      checkStatus(
+        "describeSnapshot",
+        snapshot.status.getOrElse(
+          Status(
+            errorCode = ErrorCode.UnexpectedError,
+            reason = "DescribeSnapshot Status is empty"
+          )
+        )
+      ).get
+
+      Success(
+        MilvusSnapshotInfo(
+          name = snapshot.name,
+          description = snapshot.description,
+          collectionName = snapshot.collectionName,
+          partitionNames = snapshot.partitionNames,
+          createTs = snapshot.createTs,
+          s3Location = snapshot.s3Location
+        )
+      )
+    } catch {
+      case e: StatusRuntimeException => Failure(e)
+      case e: Exception              => Failure(e)
+    }
+  }
+
   def getSegments(
       dbName: String,
       collectionName: String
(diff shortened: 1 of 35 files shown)


liliu-z:
**MilvusDataSource.scala:540** In snapshot mode, `MilvusTable.schema()` (line 348) returns the snapshot-inferred schema directly, skipping the extra-column append logic at lines 391–419. The inferred schema never includes `$segment_id`/`$row_offset`. When `pruneColumns()` appends these names from `milvus.extra.columns` (lines 524–533), line 540 calls `.find(...).get` on `None`, throwing `NoSuchElementException`. This is a guaranteed crash on the newly introduced `MilvusReadApp --mode snapshot --extra-columns` code path.

Fix: Have `MilvusTable.schema()` append extra columns in the snapshot branch before returning (mirroring lines 391–419), or change line 540 to filter out names not found in the schema.

liliu-z:
**MilvusSnapshotReader.scala:645** Three things conspire to produce nulls for V3 system fields in snapshot mode: (1) `inferSchema()` uses `toSparkSchema(includeSystemFields=true)` producing field names "RowID"/"Timestamp" from snapshot JSON; (2) `toProtobufSchemaBytes()` (line 645) strips RowID/Timestamp from protobuf schema bytes; (3) V3 reader's `fieldNameToId` (MilvusLoonPartitionReader.scala:64–65) hardcodes only lowercase `"row_id" → 0, "timestamp" → 1`. Since the protobuf schema lacks system fields, `fieldNameToId` never maps "RowID" → 0, so columns are excluded and return null. Not a regression (was null before), but the PR claims to "expose real row_id/timestamp." The V2 reader is NOT affected due to its compatible casing.

liliu-z:
**MilvusClient.scala:887** The check `message.contains("unimplemented")` on line 887 matches any exception in the cause chain containing this word, potentially triggering a silent fallback to legacy mode for unrelated errors (e.g., "Feature X is unimplemented for storage backend Y"). The gRPC status code check on lines 874–881 is correct and sufficient for standard gRPC errors. This is a plausibility-based concern rather than a proven false-positive case, but the fix is trivial.

liliu-z:
**MilvusDataSource.scala:1026** Both `MilvusDataSource.scala:1026–1042` and `MilvusReadApp.scala:292–316` read entire files into memory with no size limit. A misconfigured snapshot path pointing to a large file could cause OOM on the driver.

liliu-z:
**MilvusDataSource.scala:977** Nearly identical bucket-scoped S3A configuration code exists in `MilvusDataSource.scala:977–1024` and `MilvusReadApp.scala:318–353`. This duplication increases maintenance burden and divergence risk.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -648,56 +718,171 @@ class MilvusScan(
...
+    }
+  }
 
-    // Get collection schema and S3 config for manifest building
-    val collectionInfo = client
-      .getCollectionInfo(
-        milvusOption.databaseName,
-        milvusOption.collectionName
-      )
+  private def planInputPartitionsFromClientSnapshot(
```
liliu-z, line 931: Every client-mode read creates a server-side snapshot with 24h compaction protection via `createSnapshotForRead` (line 771) and never deletes it. The `DropSnapshot` API exists (proven by `MilvusBackfillTest.scala:122` calling `pymilvus.drop_snapshot`), but the Scala `MilvusClient` doesn't expose it. On a hot read path this accumulates hundreds of live snapshots per day per collection, each blocking segment GC.

Fix: Expose `dropSnapshot` in `MilvusClient` and call it in a `finally` block after partition planning completes (success or failure).

src/main/scala/MilvusOption.scala
```diff
@@ -65,13 +66,14 @@ object MilvusOption {
 
   val MilvusExtraColumns = "milvus.extra.columns"
   val MilvusExtraColumnPartition = "partition"
-  val MilvusExtraColumnSegmentID = "segment_id"
-  val MilvusExtraColumnRowOffset = "row_offset"
+  val MilvusExtraColumnSegmentID = "$segment_id"
```
liliu-z, line 70: Users with `milvus.extra.columns=segment_id,row_offset` silently get no extra columns after upgrade. At lines 231-236 the user's string is split and compared literally — `"segment_id" != "$segment_id"`. No error, no warning. The break propagates through all three layers: parsing here, schema appending (`appendExtraColumns` at MilvusDataSource.scala:166), and reader injection (`MilvusPartitionReaderFactory` at lines 41/91/159).

Fix: Normalize old names at parse time with a deprecation warning, mapping `segment_id` → `$segment_id` and `row_offset` → `$row_offset`.

src/main/scala/MilvusClient.scala
```diff
@@ -812,6 +866,32 @@ object MilvusClient {
...
+        case e: StatusException
+            if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
+          return true
+        case _ =>
+      }
+
+      val message = Option(current.getMessage).getOrElse("").toLowerCase
+      if (
+        message.contains(ServiceNotImplementedMarker) ||
+        message.contains("unknown method")
```
liliu-z: The `"unknown method"` substring match at line 887 walks the cause chain of **any** throwable type, not just gRPC exceptions. A network proxy returning an error body containing `"unknown method"` would silently suppress the failure and cause fallback to legacy reads. The gRPC status code check (lines 875-881) is precise; the text-based `"unknown method"` check should be narrowed to only match gRPC exception types (`StatusRuntimeException`, `StatusException`). The generic `"service not implemented"` text fallback is still needed for plain `Exception` wrappers.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -648,56 +718,171 @@ class MilvusScan(
...
+      case scala.util.Failure(e) =>
+        throw new RuntimeException(
+          s"Failed to create client read snapshot for collection " +
+            s"${milvusOption.collectionName}: ${e.getMessage}",
+          e
+        )
+    }
+  }
+
+  private def planInputPartitionsFromClientSnapshotPath(
```
liliu-z, line 1027: When `milvus.partition.id` or `milvus.segment.id` is set and the server supports `CreateSnapshot`, the connector creates a collection-wide snapshot and emits ALL segments. The user's partition/segment selector is silently ignored. The `CreateSnapshotRequest` proto is collection-scoped with no partition selector, and `planInputPartitionsFromClientSnapshotPath` (lines 806–858) delegates to `planInputPartitionsFromSnapshot` which emits all segments. The legacy path at lines 938–961 correctly enforces both selectors. A user reading a single segment on a snapshot-capable server will silently get the entire collection.

Fix: Either skip the snapshot path when partition/segment selectors are set, or post-filter the snapshot's partitions to only include matching partition/segment IDs.

src/main/scala/MilvusClient.scala
```diff
@@ -617,6 +619,58 @@ class MilvusClient(params: MilvusConnectionParams) {
     }
   }
 
+  def createSnapshotForRead(
```
liliu-z, line 625: Each read creates a uniquely-named snapshot (line 761–762) with default 86400s compaction protection, but no `DropSnapshot` call exists anywhere. The proto defines `DropSnapshot` (milvus.proto line 2629) but `MilvusClient` has no corresponding method. This means ordinary reads pin segments from compaction for 24 hours. Additionally, if `CreateSnapshot` succeeds (line 639) but `DescribeSnapshot` fails (line 641), the connector falls back to the legacy path while the orphaned snapshot remains on the server with no cleanup.

Fix: Add `dropSnapshot` to `MilvusClient`. Call it in a `finally` block after partition planning completes. For the partial-execution case, catch `DescribeSnapshot` failures separately and attempt `DropSnapshot` before falling back.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -156,6 +163,29 @@ case class MilvusTable(
     }
   logInfo(s"MilvusTable fieldIDs: $fieldIDs")
 
+  private def appendExtraColumns(schema: StructType): StructType = {
```
liliu-z, line 166: The schema now exposes `$segment_id` / `$row_offset` instead of `segment_id` / `row_offset`. The normalization at `MilvusOption.scala:209` only covers option parsing (`milvus.extra.columns`), not DataFrame column references. Existing code using `df.select("segment_id")` or SQL queries referencing the old names will break after upgrade. This should be documented in release notes with migration guidance.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -156,6 +163,29 @@ case class MilvusTable(
...
+    var fields = schema.fields.toSeq
+    def addIfRequested(name: String, field: StructField): Unit = {
+      if (
+        milvusOption.extraColumns
+          .contains(name) && !fields.exists(_.name == name)
+      ) {
+        fields = fields :+ field
+      }
+    }
+
```
liliu-z, line 176: `MilvusScanBuilder` parses `MilvusExtraColumns` without applying `normalizeExtraColumnName`, while `MilvusOption.apply()` does normalize. This causes `$segment_id`/`$row_offset` to not be filtered at line 449 (`filterNot` compares `"$segment_id"` vs `"segment_id"`), resulting in bogus synthetic field IDs via the `index + 98` heuristic at line 455. In practice readers silently drop unknown field IDs and `MilvusPartitionReaderFactory` injects correct values, so no wrong results occur. However, the synthetic ID could theoretically collide with a real user field ID.

Fix: Make `normalizeExtraColumnName` package-private and apply it in `MilvusScanBuilder`.

src/main/scala/tools/MilvusReadApp.scala
```diff
@@ -0,0 +1,407 @@
...
+    if (args.s3Bucket.isEmpty) return
+
+    MilvusOption.configureHadoopS3A(
+      conf,
+      buildStorageOptions(args),
+      args.s3Bucket
+    )
+  }
+
+  private[tools] def applyTransformations(
```
liliu-z, line 412: `applyTransformations` (lines 321-337) applies column selection first, then filter. If `--where` references columns not in `--select` (e.g. `--select id --where "tag = 'a'"`), Spark throws `AnalysisException` because `tag` has already been projected away. Standard SQL semantics apply WHERE before SELECT, and users of a CLI read tool will expect the same.

liliu-z:
**src/main/scala/operations/backfill/BackfillApp.scala:219** The check at line 219 is `i + 1 < args.length` but does not verify `!args(i + 1).startsWith("--")`. So `--mode --snapshot-path` silently accepts `"--snapshot-path"` as the value of `mode` instead of raising a missing-value error. This turns a simple CLI typo into a confusing downstream failure. In contrast, `MilvusReadApp.parseArgs` (line 102) correctly checks for this.

liliu-z:
**src/main/scala/operations/backfill/README.md:248** Lines 248-253 reference `segment_id` / `row_offset` without the `$` prefix. After this PR, the actual column names are `$segment_id` / `$row_offset`. Users following the README will get incorrect column references.

src/main/scala/tools/MilvusReadApp.scala
```diff
@@ -0,0 +1,407 @@
...
+      MilvusOption.SnapshotCollectionId -> metadata.snapshotInfo.collectionId.toString,
+      MilvusOption.SnapshotPartitionIds -> metadata.snapshotInfo.partitionIds
+        .mkString(","),
+      MilvusOption.SnapshotSchemaJson -> snapshotJson,
+      MilvusOption.SnapshotSchemaBytes -> Base64.getEncoder.encodeToString(
+        schemaBytes
+      )
+    )
+
+    metadata.storageV2ManifestList.foreach { manifestList =>
```
liliu-z: `SnapshotManifests` is only serialized when `storageV2ManifestList.nonEmpty`. For a valid empty snapshot, neither `milvus.snapshot.manifests` nor `milvus.snapshot.v2.segments` is emitted. This causes `MilvusTable.isSnapshotMode` (MilvusDataSource.scala:196) to miss the snapshot hint and fall back to client init against `dummy://snapshot-mode` instead of returning zero partitions. The snapshot planner already handles empty manifests (MilvusDataSource.scala:1033). Fix: always serialize `SnapshotManifests`, even when the list is `[]`.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -648,56 +725,180 @@ class MilvusScan(
...
+            .currentTimeMillis()}_${UUID.randomUUID().toString.replace("-", "")}"
       )
+    val description = Option(
+      options.get(MilvusOption.ClientSnapshotDescription)
+    ).filter(_.trim.nonEmpty).getOrElse("spark connector client read snapshot")
+    val protectionSeconds = Option(
+      options.get(MilvusOption.ClientSnapshotCompactionProtectionSeconds)
+    ).filter(_.trim.nonEmpty).map(_.toLong).getOrElse(86400L)
+
+    client.createSnapshotForRead(
```
liliu-z, line 951: The fast path creates a snapshot, then performs S3 metadata reads, JSON parsing, V2 segment loading, and recursive replanning. Cleanup only happens inside MilvusClient.scala:670 if `describeSnapshot` fails. Any failure after snapshot creation leaves the snapshot orphaned until TTL expiry. Since this path is now the default for normal reads, repeated transient failures will accumulate orphaned snapshots. Fix: wrap the post-creation logic in try/finally to release the snapshot on any failure.

src/main/scala/MilvusClient.scala
```diff
@@ -617,6 +620,84 @@ class MilvusClient(params: MilvusConnectionParams) {
...
+          MilvusSnapshotInfo(
+            name = snapshot.name,
+            description = snapshot.description,
+            collectionName = snapshot.collectionName,
+            partitionNames = snapshot.partitionNames,
+            createTs = snapshot.createTs,
+            s3Location = snapshot.s3Location
+          )
+        )
+      } catch {
```
liliu-z, line 661: At line 670, the inner catch calls `dropSnapshot(...)` and ignores the `Try` result. If the drop fails, there is no warning logged, making it impossible to diagnose orphaned snapshots.

Fix: Log a warning when the cleanup `dropSnapshot` call fails.

src/main/scala/MilvusClient.scala
```diff
@@ -812,6 +901,34 @@ object MilvusClient {
...
+            if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
+          return true
+        case e: StatusException
+            if e.getStatus.getCode == GrpcStatus.Code.UNIMPLEMENTED =>
+          return true
+        case _ =>
+      }
+
+      val message = Option(current.getMessage).getOrElse("").toLowerCase
+      val isGrpcException = current.isInstanceOf[StatusRuntimeException] ||
```
liliu-z: At L922, `message.contains(ServiceNotImplementedMarker)` applies to any exception, while the `unknown method` check at L924 correctly gates on `isGrpcException`. If a non-gRPC exception message happens to contain "service not implemented" (e.g. from a storage layer), the connector silently falls back to the legacy read path, potentially returning stale/inconsistent data without the user knowing.

Fix: Gate both checks on `isGrpcException`: `(isGrpcException && message.contains(ServiceNotImplementedMarker))`.

src/main/scala/MilvusOption.scala
```diff
@@ -65,13 +67,14 @@ object MilvusOption {
 
   val MilvusExtraColumns = "milvus.extra.columns"
   val MilvusExtraColumnPartition = "partition"
```
liliu-z, line 69: The extra columns were renamed from `segment_id`/`row_offset` to `$segment_id`/`$row_offset` (L69-70). `normalizeExtraColumnName` handles input config compatibility, but the output DataFrame columns now use the new names. External consumers (downstream jobs, dashboards) that reference the old column names will break silently — they'll get column-not-found errors or schema mismatches.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -543,8 +569,13 @@ object MilvusSnapshotReader {
         !includeSystemFields && (f.name == "RowID" || f.name == "Timestamp")
       )
       .map { field =>
+        val sparkFieldName = field.name match {
```
liliu-z: `toSparkSchema` (L572-576) renames system fields from `RowID`/`Timestamp` to `row_id`/`timestamp`. This affects the schema returned by `inferSchema` at `MilvusDataSource.scala` L99 when `includeSystemFields = true`. Existing snapshot-mode callers that reference `RowID`/`Timestamp` by name will break.

liliu-z:
**src/main/scala/MilvusDataSource.scala:645** `buildSnapshotHadoopConf()` (L645-659) configures per-bucket S3A settings only for the connector's bucket. If `describeSnapshot` returns a path in a different bucket, Hadoop won't find the credentials. This only affects static-credential deployments (not IAM/instance-profile) and the cross-bucket scenario is unlikely by Milvus design, but it's a defensive gap.

src/main/scala/read/MilvusPackedV2PartitionReader.scala
```diff
@@ -91,6 +91,13 @@ class MilvusPackedV2PartitionReader(
       columnGroups.flatMap(_.fieldIds.flatMap(fieldIdToName.get)).toSet
     explicitNames.filter(name => declared.contains(name)).toArray
   }
+  private val readDebugEnabled = milvusOption.options
+    .get(MilvusOption.ReaderDebug)
+    .exists(_.equalsIgnoreCase("true"))
+
+  private def debugRead(message: => String): Unit = {
+    if (readDebugEnabled) System.err.println(s"[MilvusReadDebug] $message")
```
liliu-z: All `debugRead` methods across multiple readers and `MilvusDataSource.scala` use `System.err.println`. This output cannot be controlled by log4j, lacks timestamps, and goes to stderr unconditionally when the debug flag is enabled. Not operator-friendly for production deployments.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -648,56 +742,197 @@ class MilvusScan(
...
-      .getCollectionInfo(
-        milvusOption.databaseName,
-        milvusOption.collectionName
-      )
+  private def planInputPartitionsFromClientSnapshot(
+      client: MilvusClient
+  ): Option[Array[InputPartition]] = {
+    val s3Bucket = milvusOption.options.getOrElse(
+      Properties.FsConfig.FsBucketName,
+      "a-bucket"
```
liliu-z, line 936: At lines 789 and 872, when `fs.bucket_name` is not configured, the code falls back to the literal string `"a-bucket"`. This produces confusing S3 errors (`NoSuchBucket: a-bucket`) instead of a clear configuration error message. The same fallback also exists in `Properties.scala:736`.

liliu-z:
**src/main/scala/ArrowConverter.scala:18** The `variableWidthBytes` method (lines 18-29) converts VarBinary data to Spark StringType by treating the bytes as UTF-8. If the binary data is not valid UTF-8, this will produce garbled strings. The PR intentionally fixes JSON-stored-as-binary, but there is no validation or fallback for non-UTF-8 binary data.

liliu-z:
**src/main/scala/MilvusLoonWriter.scala:658** The `setInitialCapacity(batchSize, 32.0)` call uses a fixed density of 32 bytes per value. While this fixes the previous quadratic over-allocation bug (`batchSize * 32` as total bytes), 32 bytes per value is a guess suitable for average VARCHAR fields. Columns storing long JSON blobs will experience frequent buffer reallocations.

liliu-z:
**src/main/scala/loon/Properties.scala:736** The IAM credential fix correctly defaults `accessKey`/`secretKey` to `""` instead of `"minioadmin"` when `use_iam=true`, but `FsBucketName` still defaults to `"a-bucket"`. If the bucket name is also missing in IAM mode, the error path is equally confusing.

liliu-z:
**:0** Snapshots created during reads are not released on the successful path, leading to resource leaks on the Milvus server side.

liliu-z:
**MilvusReadApp:0** MilvusReadApp.runActions can call show, count, and write.parquet on the same uncached DataFrame. Spark actions can re-plan/re-scan the source, and the client snapshot is created inside planInputPartitions, not when options are built. So one CLI command like --show 10 --count --output-parquet may create multiple server snapshots, leak all of them, and each action can observe a different collection state. The app needs either one explicit snapshot materialized into options up front, or it should persist the DataFrame before multiple actions.

liliu-z:
**:0** There is no single, consistent predicate to determine whether the connector is operating in snapshot mode. Multiple code paths check different conditions, risking divergence.

liliu-z:
**:0** The fallback path for older Milvus versions needs to be tested against a real Milvus 2.5 instance to ensure correctness.

liliu-z:
**:0** If both manifest JSON and V2 segment JSON are empty or missing, the code should fail loudly rather than returning zero partitions. Empty successful reads are dangerous and mask real problems.

liliu-z:
**:0** V3 multi-partition paths are mis-tagged. Current read correctness is unaffected because storage access uses basePath, not partitionName, but the metadata column and any code consuming it (debug logs today; backfill/routing tomorrow) will be wrong. The path-parse fix is cheap and should be done while the planner is open.

liliu-z:
**:0** The Loon writer density default should be reverted to 32.0 unless a benchmark justifies the new value.

liliu-z:
**:0** The wrapper's match-case only handles $segment_id/$row_offset; 'partition' falls through to underlying-reader column lookup, which silently returns wrong data or null. Either support it properly (the new metadata wrapper has all the info needed via p.partitionID on the partition struct) or remove it from appendExtraColumns.

liliu-z:
**:0** The dummy URI placeholder should be removed and proper CLI mode validation should be added to prevent invalid configurations.

liliu-z:
**:0** The rename of public metadata columns to $segment_id and $row_offset is a breaking change for downstream consumers and needs to be documented in release notes.

liliu-z:
**:0** When segment ID parsing fails, falling back to 0L silently produces incorrect metadata. The code should parse robustly or fail explicitly.

liliu-z:
**:0** The aliasing logic for V2 RowID and Timestamp fields deserves an actual read-path test to verify correctness.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -156,18 +162,38 @@ case class MilvusTable(
     }
   logInfo(s"MilvusTable fieldIDs: $fieldIDs")
 
+  private def appendExtraColumns(schema: StructType): StructType = {
```
liliu-z, line 166: `pruneColumns` adds `"row_id"` when Spark prunes to an empty schema (e.g. `COUNT(*)`). Snapshot schemas use `"RowID"` (uppercase from milvus-storage), so `fieldName2ID("row_id")` throws `NoSuchElementException`. Same crash in client mode when user sets `milvus.reader.fieldIDs` excluding field 0, since `fieldName2ID` won't contain `"row_id"`. Pick the first key from `fieldName2ID` instead of a magic system-field name to avoid the case-sensitivity coupling entirely.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1,11 +1,16 @@
...
+import java.net.URI
+import java.util.{Base64, HashMap, Map => JMap, UUID}
+import java.util.concurrent.atomic.AtomicBoolean
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.spark.internal.Logging
+import org.apache.spark.scheduler.{SparkListener, SparkListenerJobEnd}
```
liliu-z: The listener is global and fires on every job end once registered. A single action (e.g. `df.show()` on a partitioned scan with shuffle) can run multiple jobs, and the listener fires on the first one — dropping the snapshot while the read is still in progress. The 24-hour compaction protection masks file-deletion consequences in demos but doesn't fix correctness. Bind cleanup to the read's scope (e.g. using a query execution listener or broadcast variable destructor), not to "first job end."

src/main/scala/sources/MilvusDataSource.scala
liliu-z, line 166: `registerClientSnapshotCleanup` logs a warning and returns when no active/default `SparkSession` exists after snapshot creation, leaving the snapshot orphaned for `compactionProtectionSeconds` (default 24h). If no session is available, the code should drop the snapshot synchronously or skip creating the snapshot path entirely.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +628,143 @@ class MilvusScanBuilder(
...
+    )
+    if (v2Segments.nonEmpty) {
+      out += MilvusOption.SnapshotV2Segments ->
+        MilvusSnapshotReader.serializeV2Segments(v2Segments)
+    }
+    out
+  }
+}
+
 class MilvusScan(
```
liliu-z, line 858: If the path doesn't contain `insert_log` at all (custom layout, future variant), the method silently returns `None`, which the snapshot V3 planner then turns into either `defaultPartitionId` or an `IllegalArgumentException` mid-plan depending on how many partitionIds were supplied. Tests cover the happy path only. Consider failing explicitly when the expected path segment is missing.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +628,143 @@ class MilvusScanBuilder(
...
+      baseOptions: Map[String, String],
+      collectionName: String,
+      collectionId: Long,
+      partitionIds: Seq[Long],
+      snapshotJson: String,
+      schemaBytesBase64: String,
+      manifestList: Seq[StorageV2ManifestItem],
+      v2Segments: Seq[V2SegmentInfo]
+  ): Map[String, String] = {
+    var out = baseOptions -- SnapshotOptionKeys
```
liliu-z, line 839: `var out = baseOptions -- SnapshotOptionKeys` strips snapshot keys, but `SnapshotMode` is not in `SnapshotOptionKeys`. The subsequent `out ++ Map(SnapshotMode -> "true", ...)` sets it by last-write-wins accident, not by design. If a caller supplies `SnapshotMode=false`, the strip step keeps it, but the `++` overwrites it — correct by coincidence. Add `SnapshotMode` to `SnapshotOptionKeys` to make the intent explicit.

src/main/scala/sources/MilvusDataSource.scala
liliu-z, line 166: Self-referencing exception cause chains (where `e.getCause == e` or a cycle exists) will hang the planner in an infinite loop. Add a visited set or depth limit when walking the cause chain.

liliu-z, line 166: When snapshot drop fails (e.g. sustained describe-failure conditions), the code logs a warning but does not retry or emit a metric. Orphaned snapshots accumulate server-side. Add at least a metric counter for monitoring and consider a bounded retry.

src/main/scala/read/MilvusPackedV2PartitionReader.scala
```diff
@@ -66,15 +66,15 @@ class MilvusPackedV2PartitionReader(
...
-  // V3 readers pass a {sparkName -> fieldId-as-string} map to ArrowConverter
-  // because their on-disk parquet uses fieldId-as-string column names. V2
-  // packed parquet uses LOGICAL names (e.g. "ID"), so we must NOT remap —
-  // ArrowConverter calls `root.getVector(field.name)` directly when the
-  // mapping is empty, which is exactly what we need here. Remapping would
-  // make it look up a non-existent column and silently null every cell.
-  private val fieldNameToArrowColumn: Map[String, String] = Map.empty
+  private val fieldNameToArrowColumn: Map[String, String] = Map(
+    "row_id" -> "RowID",
+    "RowID" -> "RowID",
```
liliu-z: If a user defines a field named `row_id` or `timestamp` in their collection, `fieldNameToArrowColumn` silently reroutes it to the system column vector. This is separate from the projection filtering bug — even if projection is fixed, user-defined fields with these names will return system-column data instead of the user's data.

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -461,6 +472,30 @@ class MilvusLoonPartitionWriter(
     r.setRowCount(0)
   }
 
+  private def parsePositiveDoubleOption(
+      key: String,
+      defaultValue: Double
+  ): Double = {
+    milvusOption.options
```
liliu-z: `milvusOption.options.get(key)` where `key` is camelCase (e.g. `"milvus.writer.variableWidthBytesPerValue"`) never matches because `CaseInsensitiveStringMap.asScala` exposes keys lowercased. Only the `.orElse(milvusOption.options.get(key.toLowerCase))` branch ever matches. The first lookup is dead code that masks the intent and will silently break if `MilvusOption` is ever constructed from a non-CaseInsensitive map.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,56 +69,41 @@ class MilvusPartitionReaderFactory(
           p.readVersion
         )
 
-        // If the expected schema includes system/metadata fields, wrap the reader to add them
-        val hasRowId = schema.fieldNames.contains("row_id")
-        val hasTimestamp = schema.fieldNames.contains("timestamp")
-        val hasSegmentId = schema.fieldNames.contains("segment_id")
-        val hasRowOffset = schema.fieldNames.contains("row_offset")
+        val hasPartition = schema.fieldNames.contains(
+          MilvusOption.MilvusExtraColumnPartition
```
liliu-z, line 73: The factory has near-identical wrapper duplication between V2 and V3 branches. This already helped hide the partition semantic mismatch (now retracted). While not a correctness issue, it increases maintenance risk as another reader variant lands. Consider factoring into a shared method.

src/main/scala/sources/MilvusDataSource.scala
liliu-z, line 166: The snapshot JSON size limit is hardcoded at 64 MB. Large clusters with many segments can exceed this. Make it configurable via a Spark option with a sensible default.

src/test/scala/sources/MilvusScanClientSnapshotTest.scala
```diff
@@ -0,0 +1,269 @@
...
+
+import com.zilliz.spark.connector.read.{
+  MilvusPackedV2InputPartition,
+  MilvusSnapshotReader,
+  MilvusStorageV3InputPartition,
+  StorageV2ManifestItem,
+  V2ColumnGroup,
+  V2SegmentInfo
+}
+import com.zilliz.spark.connector.MilvusOption
```
liliu-z, line 18: The new tests cover helper plumbing but miss the dangerous runtime cases: repeated `planInputPartitions`, listener cleanup timing, V2 lowercase system column projection, snapshot `count()` with empty pruning, and user-field collision with `row_id`. The PR's highest-risk code paths (V2 packed reader system-field projection, empty-projection fallback, snapshot lifecycle) are essentially untested.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -383,6 +386,30 @@ case class SnapshotMetadata(
   */
 object MilvusSnapshotReader {
 
+  val MaxSnapshotJsonBytes: Long = 64L * 1024L * 1024L
```
liliu-z, line 390: dataTypeToSparkType (line 624) uses different numeric codes than dataTypeNameToCodeMap (line 134) for every vector type past FloatVector. The name-to-code map follows the Milvus proto enum (BinaryVector=100, Float16Vector=102, BFloat16Vector=103, SparseFloatVector=104), but dataTypeToSparkType maps 102→BinaryVector, 103→Float16Vector, 104→BFloat16Vector, 105→SparseFloatVector. This means inferSchema() returns wrong Spark types for these vector columns. The new MilvusReadApp CLI makes this latent bug user-visible. The blast radius is limited to inferSchema()'s Spark DataType mapping since toProtobufSchemaBytes uses DataType.fromValue correctly.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -90,8 +134,7 @@ object ArrowConverter extends Logging {
         vector.asInstanceOf[BitVector].get(rowIndex) != 0
 
       case StringType =>
-        val bytes = vector.asInstanceOf[VarCharVector].get(rowIndex)
-        UTF8String.fromBytes(bytes)
+        utf8StringFromVariableWidth(vector, rowIndex)
 
       case ArrayType(FloatType, _) =>
```
liliu-z, line 229: The `ArrayType(FloatType)` match at line 139 only handles the 4-byte-float-from-FixedSizeBinary decode path. `BinaryVector` and `Int8Vector` (stored as `FixedSizeBinaryVector`) hit the `ArrayType(elementType, _)` branch which casts to `ListVector`, producing a `ClassCastException`. `Float16Vector` and `BFloat16Vector` silently produce wrong float values because no half-float or bfloat16 unpacking is performed. These vector type codes (100/102/103/105) can be produced by `MilvusSnapshotReader.dataTypeCodeToSparkType` at line ~625. Either revert the Spark type mappings to `BinaryType` for unsupported codes until decoders exist, or add the decoders (half-float and bfloat16 unpack are ~10 lines each).

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,56 +69,41 @@ class MilvusPartitionReaderFactory(
...
+          MilvusOption.MilvusExtraColumnSegmentID
+        )
+        val hasRowOffset = schema.fieldNames.contains(
+          MilvusOption.MilvusExtraColumnRowOffset
+        )
 
-        if (hasRowId || hasTimestamp || hasSegmentId || hasRowOffset) {
+        if (hasPartition || hasSegmentId || hasRowOffset) {
           new PartitionReader[InternalRow] {
             private var rowOffset: Long = 0L
```
liliu-z: `$row_offset` is supposed to represent the physical segment offset, but when V3 filter pushdown is active, `MilvusLoonPartitionReader` (line ~226) skips rows internally, making `$row_offset` a post-filter row index instead. This is particularly important for the backfill workflow that depends on this column for correctness. The bug also compounds with pushed filters surviving the snapshot-mode pivot via recursive `MilvusScan` construction.

src/main/scala/MilvusOption.scala
```diff
@@ -153,6 +158,84 @@ object MilvusOption {
...
+  }
+
+  def configureHadoopS3A(
+      conf: Configuration,
+      options: Map[String, String],
+      bucket: String
+  ): Unit = {
+    val prefix = s"fs.s3a.bucket.$bucket"
+
+    conf.set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem")
```
liliu-z: The method unconditionally sets `fs.s3a.impl` globally while all other S3A settings are per-bucket. This can break environments that use a custom S3A implementation (e.g., for mocking or alternative storage backends).

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -461,6 +472,29 @@ class MilvusLoonPartitionWriter(
...
+      .filter(_.trim.nonEmpty)
+      .map { value =>
+        val parsed = Try(value.trim.toDouble).getOrElse {
+          throw new IllegalArgumentException(
+            s"$key must be a positive number, got '$value'"
+          )
+        }
+        if (parsed <= 0.0) {
+          throw new IllegalArgumentException(
+            s"$key must be a positive number, got '$value'"
```
liliu-z: The `milvus.writer.variableWidthBytesPerValue` option rejects values `<= 0` but does not reject `NaN` or `Infinity`, which would pass the positive check but produce nonsensical behavior.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -77,9 +88,7 @@ case class MilvusDataSource() extends TableProvider with DataSourceRegister {
     val milvusOption = MilvusOption(options)
 
     // Check for snapshot mode - use snapshot schema if provided
-    val isSnapshotMode =
-      Option(options.get(MilvusOption.SnapshotMode)).contains("true") ||
-        Option(options.get(MilvusOption.SnapshotManifests)).isDefined
+    val isSnapshotMode = MilvusOption.isSnapshotMode(options)
```
liliu-z, line 89: At line ~1742, a fresh `MilvusScan` is constructed with mutated options but inherits the schema and `pushedFilters` from the outer scan. If the outer scan had V3 filter pushdown applied, those filters propagate into the snapshot-mode V3 partitions. Combined with the `$row_offset` bug, filter pushdown survives the snapshot-mode pivot, producing incorrect results.

liliu-z, line 89: If `createSnapshot` succeeds but the server is briefly under load, `describeSnapshot` could return "not found." There is no retry, and `dropSnapshot` will fail because the snapshot isn't found by name. The entire sequence should ideally be a single RPC; if the proto doesn't allow that, add a bounded retry on describe.

src/main/scala/MilvusOption.scala
```diff
@@ -65,13 +67,14 @@ object MilvusOption {
 
   val MilvusExtraColumns = "milvus.extra.columns"
   val MilvusExtraColumnPartition = "partition"
-  val MilvusExtraColumnSegmentID = "segment_id"
-  val MilvusExtraColumnRowOffset = "row_offset"
+  val MilvusExtraColumnSegmentID = "$segment_id"
+  val MilvusExtraColumnRowOffset = "$row_offset"
```
liliu-z, line 71: Milvus v2.x allows user fields literally named `segment_id` or `row_offset`. The current normalization silently rewrites these to `$segment_id`/`$row_offset`, causing the user to get empty or missing columns. There is no escape hatch.

src/main/scala/tools/MilvusReadApp.scala
```diff
@@ -0,0 +1,452 @@
...
+import org.apache.spark.storage.StorageLevel
+
+import com.zilliz.spark.connector.loon.Properties
+import com.zilliz.spark.connector.read.{
+  MilvusSnapshotReader,
+  SnapshotMetadata,
+  V2SegmentInfo,
+  V2SegmentLoader
+}
+import com.zilliz.spark.connector.MilvusOption
```
liliu-z, line 19: The `MilvusOption` class defines these options, but the CLI tool (`MilvusReadApp`) doesn't expose them as arguments. The CLI tool can't actually control the new snapshot fast path it's meant to drive.

src/main/scala/tools/MilvusReadApp.scala
```diff
@@ -0,0 +1,452 @@
+package com.zilliz.spark.connector.tools
```
liliu-z, line 1: Calling `spark.sparkContext.setLogLevel("WARN")` unconditionally on a shared cluster (YARN/k8s) raises everyone's effective driver log level, hiding logs from other libraries.

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -218,6 +218,11 @@ class MilvusLoonPartitionWriter(
 
   // Batch size configuration
   private val batchSize = milvusOption.insertMaxBatchSize
+  private val variableWidthBytesPerValue =
+    parsePositiveDoubleOption(
```
liliu-z: The new option is useful but not documented anywhere. Users have no way to discover it exists or understand its behavior.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The QueryExecutionListener performs snapshot cleanup on any completed query, not just the one that created the snapshot. In a multi-query Spark session, this means one query's completion can trigger cleanup of another query's active snapshot, causing data corruption or crashes. The listener must be bound to the specific QueryExecution that created the snapshot, not registered globally.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusClient.scala:0** Setting `withDeadlineAfter` on the lazily-initialized stub creates a single absolute deadline that starts ticking at stub creation time. All subsequent RPCs share this countdown, so later RPCs get progressively less time. Long-running jobs will eventually hit deadline-exceeded errors on every call. The deadline must be set per-RPC invocation, not on the shared stub.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** After `createSnapshot` succeeds, if a subsequent operation throws an exception, there is no try-catch/finally to ensure `dropSnapshot` is called. The orphaned snapshot holds server-side resources indefinitely. This is especially dangerous with gRPC RPCs that may return ambiguous statuses — the snapshot may have been created server-side but the client doesn't know.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** When the connector detects a legacy server that doesn't support snapshots and falls back, it drops filter pushdown entirely. This is a performance regression for all users on older Milvus versions — queries that previously pushed filters down to the server now scan the full dataset on the Spark side.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/SchemaUtil.scala:0** When field IDs are missing and the code falls back to `idx + 1`, the generated IDs can collide with the hard-coded RowID (fieldID=0) and Timestamp (fieldID=1) system fields. If a user field occupies fieldID=0, the system RowID field is silently discarded, causing data misalignment at read time. Use `idx + 100` or another offset that avoids system-reserved IDs.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusOption.scala:0** MilvusOption is serialized as part of the InputPartition closure and shipped to Spark executors. This means access keys, secret keys, and tokens are embedded in the serialized task payload, which may be logged, persisted to disk, or visible in the Spark UI. Sensitive credentials must be stripped before serialization or fetched independently on the executor side.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/ArrowConverter.scala:0** ArrowConverter performs unchecked casts that will throw ClassCastException at runtime if the actual Arrow vector type doesn't match expectations. Additionally, BinaryVector values (unsigned bytes 0-255) are mapped to Spark's ByteType (signed -128 to 127), silently reinterpreting values above 127 as negative numbers. This is a silent data corruption issue.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/ArrowConverter.scala:0** Spark's BinaryType is already `byte[]`. Wrapping it in ArrayType produces `Array[byte[]]`, which doesn't correspond to any valid Milvus type. This branch is either dead code (never reached) or produces incorrect results if reached. It should be deleted.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/SchemaUtil.scala:0** SchemaUtil uses case-sensitive `existingNames.contains("RowID")` to check for system field conflicts, but MilvusLoonPartitionReader's alias mapping is case-insensitive. If a user defines a field named `row_id` (lowercase), it bypasses the guard, and the system appends a duplicate RowID field to the Arrow Schema. This can cause two fields with different IDs but semantically the same name, leading to ambiguous reads.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusClient.scala:1129** At lines 1129-1144, the retry interceptor calls `super.onClose(status, trailers)` before initiating the retry. This fires the close event on the downstream listener prematurely, meaning the application sees a failure notification and then a new RPC silently starts behind its back. This violates gRPC's listener state machine contract and can cause double-processing or incorrect error handling.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:713** When dropClientReadSnapshot fails, it retries up to 3 times in a tight loop with no Thread.sleep, backoff, or jitter. During transient network issues, these immediate retries are ineffective and add unnecessary load. For distributed cleanup operations, even a simple fixed delay would significantly improve resilience.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusPackedV2PartitionReader.scala:109** When a requested column is not present in the Column Group, the reader silently skips it. If this happens due to an upstream planning error, the query returns incomplete data with no indication of the missing column. This makes debugging extremely difficult in production.

src/main/scala/MilvusClient.scala
```diff
@@ -134,23 +139,27 @@ class MilvusClient(params: MilvusConnectionParams) {
     val server = MilvusServiceGrpc
       .blockingStub(channel)
       .withWaitForReady()
+    server
```
liliu-z, line 138: The interceptor wraps a single `ClientCall` from `next.newCall(...)` at line 1104. On retry, `executeCall()` recursively invokes `super.start(...)` on that same call from inside the listener's `onClose` (line 1157). gRPC's `ClientCall` is single-use — the second `start()` will throw `IllegalStateException("call already started")`. Combined with the removal of the pre-retry `super.onClose(...)`, the interceptor also never delivers the original failure to the caller, so the first transient UNAVAILABLE swallows the result silently. This is worse than having no retry interceptor. Either delete the custom interceptor and use gRPC's built-in `retryPolicy` via `ManagedChannelBuilder.enableRetry()` + service config, or rewrite to allocate a fresh `next.newCall(...)` per attempt.

src/test/scala/read/MilvusSnapshotReaderTest.scala
```diff
@@ -9,6 +13,52 @@ class MilvusSnapshotReaderTest extends AnyFunSuite with Matchers {
...
+        Field(name = "float16", rawDataType = Some(IntNode.valueOf(102))),
+        Field(name = "bfloat16", rawDataType = Some(IntNode.valueOf(103))),
+        Field(name = "sparse", rawDataType = Some(IntNode.valueOf(104))),
+        Field(name = "int8", rawDataType = Some(IntNode.valueOf(105)))
+      )
+    )
+
+    val sparkSchema = MilvusSnapshotReader.toSparkSchema(schema)
+
+    sparkSchema("binary").dataType shouldBe DataTypes.createArrayType(
```
liliu-z: The test at line 41 expects BinaryVector code `100` to map to `ArrayType(ByteType)`, but the implementation at `MilvusSnapshotReader.scala:625` returns `BinaryType`. The implementation matches the PR description and `DataTypeUtilTest`. This test will fail in CI. The commit `0689668` ("Address comment") appears to have flipped the expectation in the wrong direction.

src/main/scala/read/MilvusPackedV2PartitionReader.scala
```diff
@@ -24,6 +24,83 @@ import io.milvus.storage.{
...
+  private[read] case class FieldMappings(
+      fieldIdToName: Map[Long, String],
+      fieldNameToId: Map[String, Long],
+      fieldNameToArrowColumn: Map[String, String]
+  )
+
+  private[read] def buildFieldMappings(
+      milvusSchema: CollectionSchema
+  ): FieldMappings = {
+    val systemFields = Map(0L -> "RowID", 1L -> "Timestamp")
```
liliu-z, line 45: Line 37 filters lowercase aliases (`row_id`/`timestamp`) but not the CamelCase canonical names `RowID`/`Timestamp`. If a user collection has a field named `RowID` with fieldID `100`, the mapping produces `Map(0L → "RowID", 100L → "RowID")`; inverting to `fieldNameToId` is order-dependent and non-deterministic. The same issue exists in V3 at `MilvusLoonPartitionReader.scala:64`.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -62,7 +62,12 @@ class MilvusLoonPartitionReader(
   private val sourceSchema = schema
 
   private val fieldNameToId: Map[String, Long] = {
```
liliu-z: Same issue as the V2 reader: line 64 handles lowercase aliases but not the literal canonical names `RowID`/`Timestamp`. If Milvus allows those as user field names, system-field access becomes ambiguous.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -215,6 +248,7 @@ class MilvusLoonPartitionReader(
         } else {
           // Try to load next batch
           if (_currentBatch != null) {
+            _currentBatchStartRowOffset += _currentBatch.getRowCount
```
liliu-z, line 272: The vector-search branch at line 271 returns rows from `vectorSearchResults.next()` without setting `_lastReturnedRowOffset`, which defaults to `-1L`. The wrapper at `MilvusPartitionReaderFactory.scala:97` reads `underlyingReader.lastReturnedRowOffset`, so `$row_offset` is always `-1` for vector search results. Either carry the segment-local row offset alongside the row+distance pair, or reject the combination with a clear error.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,55 +69,38 @@ class MilvusPartitionReaderFactory(
...
-                resultValues(writeIdx) = rowOffset
-                rowOffset += 1
-                writeIdx += 1
+              schema.fields.zipWithIndex.foreach { case (field, writeIdx) =>
+                field.name match {
+                  case MilvusOption.MilvusExtraColumnPartition =>
+                    resultValues(writeIdx) = p.partitionName
+                  case MilvusOption.MilvusExtraColumnSegmentID =>
+                    resultValues(writeIdx) = p.segmentID
+                  case MilvusOption.MilvusExtraColumnRowOffset =>
```
liliu-z, line 97: The V3 path emits `p.partitionName` (string name like `"_default"`) while the V2 path emits `p.partitionID.toString` (numeric ID as string). A snapshot read can mix V2 + V3 segments, so users selecting the `partition` extra column get heterogeneous values from the same column depending on segment format. Both paths must agree — either both emit the partition name or both emit the ID.

liliu-z:
**src/main/scala/MilvusDataSource.scala:0** Snapshot cleanup is tied to `SparkListenerSQLExecutionEnd` only. SIGKILL, OOM, or `yarn application -kill` leaves orphan snapshot records on the Milvus server. The driver-death case cannot be fully solved in Spark client code, but the PR needs either a documented Milvus-side TTL/sweeper guarantee or a user-visible warning that snapshots may accumulate after unclean driver exits.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,55 +69,38 @@ class MilvusPartitionReaderFactory(
           p.readVersion
         )
 
-        // If the expected schema includes system/metadata fields, wrap the reader to add them
-        val hasRowId = schema.fieldNames.contains("row_id")
-        val hasTimestamp = schema.fieldNames.contains("timestamp")
-        val hasSegmentId = schema.fieldNames.contains("segment_id")
-        val hasRowOffset = schema.fieldNames.contains("row_offset")
+        val hasPartition = schema.fieldNames.contains(
+          MilvusOption.MilvusExtraColumnPartition
```
liliu-z, line 73: The V3 wrapper reads `underlyingReader.lastReturnedRowOffset` (real segment-relative offset accounting for filter-skipped rows). The V2 wrapper uses a wrapper-local `var rowOffset` counter that increments per `get()` — a post-filter sequential index, not the segment-relative offset. If downstream code uses `$row_offset` to re-order rows for backfill (the documented use case), V2-segment rows will be misordered relative to V3-segment rows in the same collection.

src/main/scala/MilvusClient.scala
```diff
@@ -134,23 +139,27 @@ class MilvusClient(params: MilvusConnectionParams) {
...
+              localTime = java.time.LocalDateTime.now().toString,
+              host = java.net.InetAddress.getLocalHost.getHostName,
+              user = "scala-sdk-user"
+            )
           )
         )
       )
-    )
     server
   }
```
liliu-z, line 155: The fallback that checks for substring matches in UNKNOWN-status error messages is fragile across proxy versions. It would not match if the message is ever rephrased (e.g., "method not registered" instead of the current wording).

liliu-z:
**src/main/scala/MilvusReadApp.scala:0** Passing S3 credentials via `--s3-access-key` and `--s3-secret-key` CLI arguments makes them visible to every process on the host via `ps` or `/proc`. The CLI app should also accept `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` environment variables or a credentials file as alternatives. At minimum, document "use --use-iam in production" prominently.

src/main/scala/MilvusOption.scala
```diff
@@ -153,6 +158,72 @@ object MilvusOption {
     "milvus.snapshot.schema.json" // Optional: raw schema JSON for building MilvusCollectionInfo
   val SnapshotSchemaBytes =
     "milvus.snapshot.schema.bytes" // Base64 encoded protobuf CollectionSchema bytes
+  val SnapshotMaxJsonBytes = "milvus.snapshot.maxJsonBytes"
```
liliu-z, line 161: Sets `fs.s3a.bucket.$bucket.path.style.access=true` and `fs.s3a.bucket.$bucket.connection.ssl.enabled=false` unconditionally, overriding any explicit global Hadoop config the user expects to inherit. The `fs.s3a.impl` check preserves existing values; these two keys should follow the same pattern or be configurable.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -821,6 +1254,27 @@ class MilvusScan(
...
+            case Left(e) =>
+              throw new Exception(
+                s"Failed to parse SnapshotV2Segments: ${e.getMessage}",
+                e
+              )
+          }
+        }
+        .getOrElse(Seq.empty)
+
+    if (manifestList.isEmpty && packedV2Segments.isEmpty) {
```
liliu-z: A valid empty collection triggers `IllegalArgumentException("Snapshot mode has no StorageV3 manifests or StorageV2 segments")`. This should return an empty DataFrame (`Array.empty[InputPartition]`). The guard is redundant for safety since `parseSnapshotMetadata` already fails loudly on unparseable data. Distinguish "zero segments" (return empty) from "unparseable snapshot" (fail loud) and remove this overly aggressive guard.

liliu-z:
**src/main/scala/MilvusPartitionReaderFactory.scala:94** The `partition` extra column is emitted as a raw Java `String` instead of `UTF8String.fromString(...)`. Spark's `InternalRow` contract requires UTF8String for string-type columns; a raw String will cause a ClassCastException or silent data corruption at read time. The same issue exists in the V3 wrapper at line 168. Wrap the value with `UTF8String.fromString(...)`.

liliu-z:
**src/main/scala/MilvusDataSource.scala:384** When the Milvus collection already contains fields named `row_id` or `timestamp`, `MilvusDataSource.schema()` (lines 384-410) will produce a schema with duplicate column names — one from the system fields and one from the user fields. This causes ambiguous column resolution at query time. Apply the same conflict/dedup policy that `SchemaUtil` already uses on the Arrow side.

liliu-z:
**src/main/scala/MilvusScanBuilder.scala:0** The `pruneColumns` implementation derives field IDs positionally (`index < 2 → fieldID = 0/1`), which assumes system fields (RowID, Timestamp) are always the first two columns. When an external schema omits system fields, the reader will silently project RowID/Timestamp instead of the requested user fields, returning wrong column data. Source field IDs from `milvusCollection.schema.fields(...).fieldID` instead of position.

src/main/scala/MilvusOption.scala
```diff
@@ -153,6 +158,103 @@ object MilvusOption {
...
+  val ClientSnapshotName = "milvus.client.snapshot.name"
+  val ClientSnapshotDescription = "milvus.client.snapshot.description"
+  val ClientSnapshotCompactionProtectionSeconds =
+    "milvus.client.snapshot.compactionProtectionSeconds"
+
+  def isSnapshotMode(options: Map[String, String]): Boolean = {
+    options.get(SnapshotMode).exists(_.equalsIgnoreCase("true")) ||
+    options.contains(SnapshotManifests) ||
```
liliu-z, line 169: Both overloads of `isSnapshotMode` use `.contains` to check whether the snapshot manifest key is present. Setting `milvus.snapshot.manifests=""` will trigger snapshot mode even though no manifests are specified, which is fragile. Filter out empty/blank values before treating the option as present.

src/main/scala/MilvusClient.scala
```diff
@@ -617,12 +622,129 @@ class MilvusClient(params: MilvusConnectionParams) {
...
+        return snapshot
+      } catch {
+        case e: Exception =>
+          lastFailure = Some(e)
+          if (attempt < maxAttempts) {
+            logWarning(
+              s"describeSnapshot failed for $snapshotName (attempt $attempt/$maxAttempts)",
+              e
+            )
+            TimeUnit.MILLISECONDS.sleep(200L * attempt)
```
liliu-z, line 713: When an `InterruptedException` is caught during the retry sleep at line 713, the interrupt flag is cleared. The method then throws without restoring `Thread.currentThread().interrupt()`, which violates the Java interrupt contract and can cause callers to miss the interrupt signal. Catch `InterruptedException`, restore the interrupt flag, and stop retrying.

src/main/scala/MilvusClient.scala
```diff
@@ -83,10 +87,7 @@ class MilvusClient(params: MilvusConnectionParams) {
       }
     }
 
-    val interceptors = Seq(
-      getConnectionMetadataInterceptor(),
-      retryInterceptor
-    )
+    val interceptors = Seq(getConnectionMetadataInterceptor())
```
liliu-z, line 90: The gRPC channel builder calls `.enableRetry()` but no `defaultServiceConfig` is installed, so the retry policy is never actually applied. Either install a proper service config with retry parameters, or remove the misleading `.enableRetry()` call.

LawrenceTL92:
Closing: the content of this PR has landed on main through the split follow-ups — snapshot RPC client (#95), client snapshot planning lifecycle (#96), canonical `$segment_id`/`$row_offset` metadata columns (#97), RowID/Timestamp alias unification (#98) and hardened Arrow decoding (#99). The only piece not on main is the `MilvusReadApp` CLI wrapper; if it is still wanted, please open a fresh, focused PR against current main.

PR #95, opened 2026-05-21
enhance: [reader] add snapshot RPC client support

congqixia:
Add Milvus snapshot create/describe/drop wrappers and a bounded gRPC fallback classifier so the datasource split can depend on a small client-side contract first.


liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusClient.scala:652** `checkStatus("createSnapshot", createStatus).get` deliberately unwraps a `Try`, throws on failure, and the outer `catch` block re-wraps it as `Failure(e)`. This is a Scala anti-pattern — using exceptions for control flow when you already hold a `Try`. It's slower, harder to read, and obscures the actual error-handling intent.

PR #96, opened 2026-05-25
enhance: [reader] add client snapshot planning lifecycle

congqixia:
Use snapshot RPCs as the default client-read planning path when no partition or segment selector is set, then reuse snapshot-only planning from the returned metadata. Fall back to legacy segment planning only when snapshot RPCs are unavailable, and add bounded snapshot JSON reads plus focused planner tests.

### DIFF
--- a/src/main/scala/read/MilvusSnapshotReader.scala
+++ b/src/main/scala/read/MilvusSnapshotReader.scala
@@ -383,6 +383,30 @@ case class SnapshotMetadata(
   */
 object MilvusSnapshotReader {
 
+  val MaxSnapshotJsonBytes: Long = 64L * 1024L * 1024L
+
+  def readUtf8WithLimit(
+      in: java.io.InputStream,
+      path: String,
+      maxBytes: Long = MaxSnapshotJsonBytes
+  ): String = {
+    val out = new java.io.ByteArrayOutputStream()
+    val buf = new Array[Byte](8192)
+    var total = 0L
+    var n = in.read(buf)
+    while (n >= 0) {
+      total += n
+      if (total > maxBytes) {
+        throw new IllegalArgumentException(
+          s"Snapshot metadata $path exceeds max size $maxBytes bytes"
+        )
+      }
+      out.write(buf, 0, n)
+      n = in.read(buf)
+    }
+    out.toString(java.nio.charset.StandardCharsets.UTF_8.name())
+  }
+
   private val mapper: ObjectMapper with ScalaObjectMapper = {
     val m = new ObjectMapper() with ScalaObjectMapper
     m.registerModule(DefaultScalaModule)
@@ -428,12 +452,11 @@ object MilvusSnapshotReader {
       path: String
   ): Either[Throwable, SnapshotMetadata] = {
     try {
-      val source = scala.io.Source.fromFile(path)
+      val in = new java.io.FileInputStream(path)
       try {
-        val json = source.mkString
-        parseSnapshotMetadata(json)
+        parseSnapshotMetadata(readUtf8WithLimit(in, path))
       } finally {
-        source.close()
+        in.close()
       }
     } catch {
       case e: Exception => Left(e)
--- a/src/main/scala/read/MilvusSnapshotReader.scala
+++ b/src/main/scala/read/MilvusSnapshotReader.scala
@@ -390,6 +390,11 @@ object MilvusSnapshotReader {
       path: String,
       maxBytes: Long = MaxSnapshotJsonBytes
   ): String = {
+    if (maxBytes <= 0) {
+      throw new IllegalArgumentException(
+        s"Snapshot metadata max size must be positive, got $maxBytes"
+      )
+    }
     val out = new java.io.ByteArrayOutputStream()
     val buf = new Array[Byte](8192)
     var total = 0L
--- a/src/main/scala/read/MilvusSnapshotReader.scala
+++ b/src/main/scala/read/MilvusSnapshotReader.scala
@@ -14,6 +14,8 @@ import com.fasterxml.jackson.module.scala.{
 }
 import org.apache.spark.sql.types._
 
+import com.zilliz.spark.connector.serde.ArrowConverter
+
 /** Helper object for converting JSON values that may be either numeric or
   * string to Long/Int Milvus snapshot JSON format may serialize numbers as
   * strings in some versions
@@ -687,7 +689,10 @@ object MilvusSnapshotReader {
         StructField(
           field.name,
           dataTypeToSparkType(field.dataType, field.elementType),
-          nullable = field.nullable.getOrElse(true)
+          nullable = field.nullable.getOrElse(true),
+          metadata = new MetadataBuilder()
+            .putLong(ArrowConverter.MilvusDataTypeMetadataKey, field.dataType)
+            .build()
         )
       }
     StructType(userFields)
--- a/src/main/scala/read/MilvusSnapshotReader.scala
+++ b/src/main/scala/read/MilvusSnapshotReader.scala
@@ -81,6 +81,23 @@ private[read] object JsonTypeConverter {
       }
     case _ => 0
   }
+
+  def toFieldStateCode(value: Any): Int = value match {
+    case i: Int    => i
+    case l: Long   => l.toInt
+    case n: Number => n.intValue()
+    case s: String =>
+      try s.toInt
+      catch { case _: NumberFormatException => Field.fieldStateNameToCode(s) }
+    case node: JsonNode if node.isNumber => node.asInt()
+    case node: JsonNode if node.isTextual =>
+      val text = node.asText()
+      try text.toInt
+      catch {
+        case _: NumberFormatException => Field.fieldStateNameToCode(text)
+      }
+    case _ => 0
+  }
 }
 
 /** Type parameter for Milvus field
@@ -101,7 +118,18 @@ case class Field(
       None, // Can be Int or String (e.g., 5 or "Int64")
     @JsonProperty("is_primary_key") isPrimaryKey: Option[Boolean] = None,
     @JsonProperty("is_clustering_key") isClusteringKey: Option[Boolean] = None,
-    @JsonProperty("type_params") typeParams: Option[Seq[TypeParam]] = None
+    @JsonProperty("type_params") typeParams: Option[Seq[TypeParam]] = None,
+    @JsonProperty("autoID") @JsonAlias(Array("auto_id")) autoID: Option[
+      Boolean
+    ] = None,
+    @JsonProperty("state") rawState: Option[JsonNode] = None,
+    @JsonProperty("element_type") rawElementType: Option[JsonNode] = None,
+    @JsonProperty("is_dynamic") isDynamic: Option[Boolean] = None,
+    @JsonProperty("is_partition_key") isPartitionKey: Option[Boolean] = None,
+    @JsonProperty("nullable") nullable: Option[Boolean] = None,
+    @JsonProperty("is_function_output") isFunctionOutput: Option[Boolean] =
+      None,
+    @JsonProperty("default_value") defaultValue: Option[JsonNode] = None
 ) {
 
   /** Get the data type as Int, handling both numeric and string formats String
@@ -121,7 +149,15 @@ case class Field(
       case _          => 0L
     }
   }
+
+  def state: Int = rawState.map(JsonTypeConverter.toFieldStateCode).getOrElse(0)
+
+  def elementType: Int =
+    rawElementType.map(JsonTypeConverter.toDataTypeCode).getOrElse(0)
 }
+
+case class UnsupportedSnapshotSchemaField(message: String)
+    extends IllegalArgumentException(message)
 
 object Field {
 
@@ -139,27 +175,36 @@ object Field {
     "Double" -> 11,
     "String" -> 20,
     "VarChar" -> 21,
-    "JSON" -> 22,
-    "Array" -> 23,
-    // Array element types (23-30 based on element type)
-    "ArrayBool" -> 23,
-    "ArrayInt8" -> 24,
-    "ArrayInt16" -> 25,
-    "ArrayInt32" -> 26,
-    "ArrayInt64" -> 27,
-    "ArrayFloat" -> 28,
-    "ArrayDouble" -> 29,
-    "ArrayVarChar" -> 30,
+    "Array" -> 22,
+    "JSON" -> 23,
+    "Geometry" -> 24,
+    "Text" -> 25,
+    "Timestamptz" -> 26,
     // Vector types
     "BinaryVector" -> 100,
     "FloatVector" -> 101,
     "Float16Vector" -> 102,
     "BFloat16Vector" -> 103,
-    "SparseFloatVector" -> 104
+    "SparseFloatVector" -> 104,
+    "Int8Vector" -> 105,
+    "ArrayOfVector" -> 106,
+    "ArrayOfStruct" -> 200,
+    "Struct" -> 201
+  )
+
+  private val fieldStateNameToCodeMap: Map[String, Int] = Map(
+    "FieldCreated" -> 0,
+    "FieldCreating" -> 1,
+    "FieldDropping" -> 2,
+    "FieldDropped" -> 3
   )
 
   def dataTypeNameToCode(name: String): Int = {
     dataTypeNameToCodeMap.getOrElse(name, 0)
+  }
+
+  def fieldStateNameToCode(name: String): Int = {
+    fieldStateNameToCodeMap.getOrElse(name, 0)
   }
 }
 
@@ -176,7 +221,12 @@ case class CollectionSchema(
     @JsonProperty("name") name: String,
     @JsonProperty("description") description: Option[String] = None,
     @JsonProperty("fields") fields: Seq[Field],
-    @JsonProperty("properties") properties: Option[Seq[Property]] = None
+    @JsonProperty("properties") properties: Option[Seq[Property]] = None,
+    @JsonProperty("autoID") @JsonAlias(Array("auto_id")) autoID: Option[
+      Boolean
+    ] = None,
+    @JsonProperty("enable_dynamic_field") enableDynamicField: Option[Boolean] =
+      None
 ) {
   def getFieldByName(name: String): Option[Field] = {
     fields.find(_.name == name)
@@ -429,6 +479,69 @@ object MilvusSnapshotReader {
     m
   }
 
+  private def protobufDefaultValue(
+      node: JsonNode,
+      fieldName: String
+  ): io.milvus.grpc.schema.ValueField = {
+    import io.milvus.grpc.schema.ValueField
+
+    val value = Option(node).filterNot(_.isNull).getOrElse {
+      throw UnsupportedSnapshotSchemaField(
+        s"default_value for field $fieldName is null"
+      )
+    }
+
+    def child(names: String*): Option[JsonNode] =
+      names.iterator.map(value.get).find(v => v != null && !v.isNull)
+
+    child("bool_data", "boolData")
+      .map(v => ValueField().withBoolData(v.asBoolean()))
+      .orElse(
+        child("int_data", "intData").map(v =>
+          ValueField().withIntData(v.asInt())
+        )
+      )
+      .orElse(
+        child("long_data", "longData").map(v =>
+          ValueField().withLongData(v.asLong())
+        )
+      )
+      .orElse(
+        child("float_data", "floatData").map(v =>
+          ValueField().withFloatData(v.floatValue())
+        )
+      )
+      .orElse(
+        child("double_data", "doubleData").map(v =>
+          ValueField().withDoubleData(v.asDouble())
+        )
+      )
+      .orElse(
+        child("string_data", "stringData").map(v =>
+          ValueField().withStringData(v.asText())
+        )
+      )
+      .orElse(
+        child("bytes_data", "bytesData").map { v =>
+          ValueField().withBytesData(
+            com.google.protobuf.ByteString.copyFrom(
+              java.util.Base64.getDecoder.decode(v.asText())
+            )
+          )
+        }
+      )
+      .orElse(
+        child("timestamptz_data", "timestamptzData").map(v =>
+          ValueField().withTimestamptzData(v.asLong())
+        )
+      )
+      .getOrElse {
+        throw UnsupportedSnapshotSchemaField(
+          s"default_value for field $fieldName uses an unsupported shape: $value"
+        )
+      }
+  }
+
   /** Parse snapshot metadata from JSON string
     *
     * @param json
@@ -573,8 +686,8 @@ object MilvusSnapshotReader {
       .map { field =>
         StructField(
           field.name,
-          dataTypeToSparkType(field.dataType, field.typeParams),
-          nullable = true
+          dataTypeToSparkType(field.dataType, field.elementType),
+          nullable = field.nullable.getOrElse(true)
         )
       }
     StructType(userFields)
@@ -588,7 +701,7 @@ object MilvusSnapshotReader {
     *   Corresponding Spark DataType
     */
   def fieldToSparkType(field: Field): DataType = {
-    dataTypeToSparkType(field.dataType, field.typeParams)
+    dataTypeToSparkType(field.dataType, field.elementType)
   }
 
   /** Convert Milvus data type to Spark DataType
@@ -602,7 +715,7 @@ object MilvusSnapshotReader {
     */
   private def dataTypeToSparkType(
       dataType: Int,
-      typeParams: Option[Seq[TypeParam]]
+      elementType: Int
   ): DataType = {
     dataType match {
       case 1   => BooleanType // Bool
@@ -614,20 +727,17 @@ object MilvusSnapshotReader {
       case 11  => DoubleType // Double
       case 20  => StringType // String
       case 21  => StringType // VarChar
-      case 22  => StringType // JSON (as string)
-      case 23  => ArrayType(BooleanType) // Array[Bool]
-      case 24  => ArrayType(ByteType) // Array[Int8]
-      case 25  => ArrayType(ShortType) // Array[Int16]
-      case 26  => ArrayType(IntegerType) // Array[Int32]
-      case 27  => ArrayType(LongType) // Array[Int64]
-      case 28  => ArrayType(FloatType) // Array[Float]
-      case 29  => ArrayType(DoubleType) // Array[Double]
-      case 30  => ArrayType(StringType) // Array[VarChar]
+      case 22  => ArrayType(dataTypeToSparkType(elementType, 0)) // Array
+      case 23  => StringType // JSON (as string)
+      case 24  => StringType // Geometry
+      case 25  => StringType // Text
+      case 26  => LongType // Timestamptz
+      case 100 => ArrayType(ByteType) // BinaryVector
       case 101 => ArrayType(FloatType) // FloatVector
-      case 102 => ArrayType(ByteType) // BinaryVector
-      case 103 => ArrayType(ShortType) // Float16Vector
-      case 104 => ArrayType(ShortType) // BFloat16Vector
-      case 105 => MapType(LongType, FloatType) // SparseFloatVector
+      case 102 => ArrayType(FloatType) // Float16Vector
+      case 103 => ArrayType(FloatType) // BFloat16Vector
+      case 104 => MapType(LongType, FloatType) // SparseFloatVector
+      case 105 => ArrayType(ByteType) // Int8Vector
       case _   => BinaryType // Unknown types as binary
     }
   }
@@ -665,8 +775,9 @@ object MilvusSnapshotReader {
   def toProtobufSchemaBytes(schema: CollectionSchema): Array[Byte] = {
     import io.milvus.grpc.schema.{
       CollectionSchema => ProtoCollectionSchema,
+      DataType,
       FieldSchema,
-      DataType
+      FieldState
     }
     import io.milvus.grpc.common.KeyValuePair
 
@@ -684,14 +795,28 @@ object MilvusSnapshotReader {
         isClusteringKey = field.isClusteringKey.getOrElse(false),
         typeParams = field.typeParams.getOrElse(Seq.empty).map { tp =>
           KeyValuePair(key = tp.key, value = tp.value)
-        }
+        },
+        autoID = field.autoID.getOrElse(false),
+        state = FieldState.fromValue(field.state),
+        elementType = DataType.fromValue(field.elementType),
+        isDynamic = field.isDynamic.getOrElse(false),
+        isPartitionKey = field.isPartitionKey.getOrElse(false),
+        nullable = field.nullable.getOrElse(false),
+        isFunctionOutput = field.isFunctionOutput.getOrElse(false),
+        defaultValue =
+          field.defaultValue.map(protobufDefaultValue(_, field.name))
       )
     }
 
     val protoSchema = ProtoCollectionSchema(
       name = schema.name,
       description = schema.description.getOrElse(""),
-      fields = protoFields
+      autoID = schema.autoID.getOrElse(false),
+      fields = protoFields,
+      enableDynamicField = schema.enableDynamicField.getOrElse(false),
+      properties = schema.properties.getOrElse(Seq.empty).map { prop =>
+        KeyValuePair(key = prop.key, value = prop.value)
+      }
     )
 
     protoSchema.toByteArray
--- a/src/main/scala/serde/ArrowConverter.scala
+++ b/src/main/scala/serde/ArrowConverter.scala
@@ -78,14 +78,21 @@ object ArrowConverter extends Logging {
       rowIndex: Int,
       sparkType: DataType
   ): Any = {
-    arrowValueToSparkValue(vector, rowIndex, sparkType, None)
+    arrowValueToSparkValue(
+      vector,
+      rowIndex,
+      sparkType,
+      None,
+      allowLegacyFloatVector = true
+    )
   }
 
   private def arrowValueToSparkValue(
       vector: FieldVector,
       rowIndex: Int,
       sparkType: DataType,
-      milvusType: Option[MilvusDataType]
+      milvusType: Option[MilvusDataType],
+      allowLegacyFloatVector: Boolean = false
   ): Any = {
     sparkType match {
       case LongType =>
@@ -123,7 +130,13 @@ object ArrowConverter extends Logging {
       case ArrayType(FloatType, _)
           if vector.isInstanceOf[FixedSizeBinaryVector] =>
         val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
-        ArrayData.toArrayData(decodeFixedSizeBinaryFloats(bytes, milvusType))
+        ArrayData.toArrayData(
+          decodeFixedSizeBinaryFloats(
+            bytes,
+            milvusType,
+            allowLegacyFloatVector
+          )
+        )
 
       case ArrayType(FloatType, _) =>
         val listVector = vector.asInstanceOf[ListVector]
@@ -204,7 +217,8 @@ object ArrowConverter extends Logging {
 
   private def decodeFixedSizeBinaryFloats(
       bytes: Array[Byte],
-      milvusType: Option[MilvusDataType]
+      milvusType: Option[MilvusDataType],
+      allowLegacyFloatVector: Boolean
   ): Array[Float] = {
     milvusType match {
       case Some(MilvusDataType.Float16Vector) =>
@@ -217,10 +231,24 @@ object ArrowConverter extends Logging {
           .grouped(2)
           .map(b => FloatConverter.fromBFloat16Bytes(b.toSeq))
           .toArray
-      case _ =>
-        val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
-        (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
+      case Some(MilvusDataType.FloatVector) =>
+        decodeFloatVectorBytes(bytes)
+      case None if allowLegacyFloatVector =>
+        decodeFloatVectorBytes(bytes)
+      case None =>
+        throw new IllegalArgumentException(
+          s"FixedSizeBinary float vectors require ${MilvusDataTypeMetadataKey} metadata"
+        )
+      case Some(other) =>
+        throw new IllegalArgumentException(
+          s"Cannot decode FixedSizeBinary as Array[Float] for Milvus type $other"
+        )
     }
+  }
+
+  private def decodeFloatVectorBytes(bytes: Array[Byte]): Array[Float] = {
+    val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
+    (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
   }
 
   val MilvusDataTypeMetadataKey = "milvus.data_type"
--- a/src/main/scala/serde/ArrowConverter.scala
+++ b/src/main/scala/serde/ArrowConverter.scala
@@ -11,6 +11,9 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 
+import com.zilliz.spark.connector.FloatConverter
+import io.milvus.grpc.schema.{DataType => MilvusDataType}
+
 /** Utilities for converting between Spark InternalRow and Arrow vectors
   */
 object ArrowConverter extends Logging {
@@ -47,7 +50,12 @@ object ArrowConverter extends Logging {
       } else if (vector.isNull(rowIndex)) {
         values(index) = null
       } else {
-        values(index) = arrowValueToSparkValue(vector, rowIndex, field.dataType)
+        values(index) = arrowValueToSparkValue(
+          vector,
+          rowIndex,
+          field.dataType,
+          milvusDataType(field)
+        )
       }
     }
 
@@ -69,6 +77,15 @@ object ArrowConverter extends Logging {
       vector: FieldVector,
       rowIndex: Int,
       sparkType: DataType
+  ): Any = {
+    arrowValueToSparkValue(vector, rowIndex, sparkType, None)
+  }
+
+  private def arrowValueToSparkValue(
+      vector: FieldVector,
+      rowIndex: Int,
+      sparkType: DataType,
+      milvusType: Option[MilvusDataType]
   ): Any = {
     sparkType match {
       case LongType =>
@@ -93,13 +110,33 @@ object ArrowConverter extends Logging {
         val bytes = vector.asInstanceOf[VarCharVector].get(rowIndex)
         UTF8String.fromBytes(bytes)
 
-      case ArrayType(FloatType, _) =>
-        // FloatVector stored as FixedSizeBinaryVector
+      case ArrayType(ByteType, _)
+          if vector.isInstanceOf[FixedSizeBinaryVector] =>
         val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
-        val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
-        val floats =
-          (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
-        ArrayData.toArrayData(floats)
+        ArrayData.toArrayData(bytes)
+
+      case ArrayType(ShortType, _)
+          if vector.isInstanceOf[FixedSizeBinaryVector] =>
+        val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
+        ArrayData.toArrayData(bytes.map(_.toShort))
+
+      case ArrayType(FloatType, _)
+          if vector.isInstanceOf[FixedSizeBinaryVector] =>
+        val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
+        ArrayData.toArrayData(decodeFixedSizeBinaryFloats(bytes, milvusType))
+
+      case ArrayType(FloatType, _) =>
+        val listVector = vector.asInstanceOf[ListVector]
+        val dataVector = listVector.getDataVector
+        val startIndex = listVector.getElementStartIndex(rowIndex)
+        val endIndex = listVector.getElementEndIndex(rowIndex)
+        val length = endIndex - startIndex
+        val arrayElements = (0 until length).map { i =>
+          val elemIndex = startIndex + i
+          if (dataVector.isNull(elemIndex)) null
+          else arrowValueToSparkValue(dataVector, elemIndex, FloatType, None)
+        }.toArray
+        ArrayData.toArrayData(arrayElements)
 
       case ArrayType(elementType, _) =>
         // Generic array handling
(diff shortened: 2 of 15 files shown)


liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** planInputPartitionsFromClientSnapshot creates a server-side snapshot, fetches up to 64 MB of JSON from S3, parses manifests, and recursively plans partitions BEFORE checking whether SparkSession and executionId are available for cleanup registration. If cleanup cannot be registered (DataFrame/RDD API, custom schedulers), all that work is discarded, the snapshot is dropped, and the code falls back to the legacy path. Fix: check SparkSession.getActiveSession and spark.sql.execution.id BEFORE calling createSnapshotForRead. Place the probe in the instance method, not in the pure canUseClientSnapshotFastPath helper, to keep that helper testable without faking a SparkContext.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusSnapshotReader.scala:0** buildSnapshotHadoopConf constructs a bare Configuration instead of basing it on SparkSession.active.sessionState.newHadoopConf(). This loses all session-injected credentials (IAM roles, KMS keys, custom S3 endpoints, Azure/GCS tokens). Any cloud-deployed Spark cluster relying on session-scoped credentials will fail to read snapshot files from object storage. Fix: use SparkSession.active.sessionState.newHadoopConf() as the base configuration.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** buildClientSnapshotOptions includes the full SnapshotSchemaJson (up to 64 MB) in the options map. This map is captured in the recursive MilvusScan constructor and serialized into every task closure. With N partitions this becomes N × 64 MB of serialized data on the Driver, causing immediate OOM. Fix: remove SnapshotSchemaJson from buildClientSnapshotOptions entirely — the schema is already available through MilvusScan's own schema field.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The SparkListener onOtherEvent callback performs synchronous MilvusClient construction and dropClientReadSnapshot RPC calls directly on the LiveListenerBus thread. If the Milvus server is slow or unreachable, this blocks the entire event bus, stalling job scheduling, stage transitions, and UI updates for all Spark jobs in the cluster. Fix: wrap cleanup operations in a Future on a dedicated single-thread daemon ExecutionContext or use ExecutionContext.global.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The AtomicBoolean guard in the cleanup callback is set (compareAndSet) before the MilvusClient is constructed and the drop RPC is attempted. If client construction fails, the boolean is already true and no retry will ever be attempted, leaving the server-side snapshot as a permanent orphan. Additionally, dropClientReadSnapshot returns Boolean, swallowing the actual error cause. Fix: flip the AtomicBoolean only after the drop RPC completes successfully, and change dropClientReadSnapshot to return Try/Either or throw to preserve error provenance.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** Each snapshot-based scan registers a new SparkListener via addSparkListener. Under a long-lived SparkContext (e.g., Thrift Server), listeners accumulate indefinitely with no TTL or deregistration. Worse, each cleanup callback constructs a fresh MilvusClient (new gRPC channel + TCP connection), sends one RPC, and immediately closes. At high query concurrency (e.g., 100 short queries/sec), this creates a connection storm that exhausts ephemeral ports and file descriptors (TIME_WAIT accumulation) on the Driver and hammers the Milvus server with connection handshakes. Fix: add a TTL-based Listener eviction mechanism, and use a shared/pooled MilvusClient or a daemon queue that batches drop requests over a single long-lived connection.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** When cleanup registration fails and the code attempts to drop the just-created snapshot before falling back to legacy, if dropClientReadSnapshot also fails (e.g., network flap, 3 retries exhausted returning false), the code throws IllegalStateException, killing the entire Spark job. The intent was graceful degradation to legacy path, but a secondary cleanup failure prevents degradation entirely. The server has a 24-hour TTL safety net for orphaned snapshots. Fix: catch the drop failure, log at ERROR level, and unconditionally return None to proceed with legacy fallback. Never let auxiliary snapshot cleanup block the primary data read path.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** On the planning-failure path, drop failure is swallowed (boolean return). On the cleanup-registration-failure path, drop failure throws IllegalStateException. The error handling is inconsistent and neither path preserves the root cause. Fix: unify via addSuppressed — on planning failure, if drop also fails, add the drop error as a suppressed exception on the original error before rethrowing. On registration failure, log and degrade rather than throw. Requires dropClientReadSnapshot to expose the underlying exception rather than flattening to Boolean.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusSnapshotReader.scala:0** readUtf8WithLimit has no lower-bound validation on the maxBytes parameter. If a user sets milvus.snapshot.maxJsonBytes to -1 thinking it means 'unlimited', the first read of 8192 bytes immediately exceeds -1 and throws a confusing 'exceeds max size' error. Fix: treat maxBytes <= 0 as 'use default' or fail loudly at configuration parse time with a clear message.

liliu-z:
**MilvusScan.scala:940** `getCollectionInfo` is called unconditionally before checking `canUseClientSnapshotFastPath`. On the fast path, the snapshot JSON already contains the full schema, so the RPC result is never used. This adds unnecessary network latency to every query that takes the fast path.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** In `planInputPartitionsFromClientSnapshot`, `resolveClientSnapshotLocation` is called outside the `try` block. If it throws (e.g., empty s3Location, invalid scheme), the already-created snapshot is never cleaned up by the `catch` block's `dropClientReadSnapshot`. The `val snapshotPath = ...` line must be moved inside the `try` block.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** When `SparkContext.stop()` fires, the listener submits a Future for `dropClientReadSnapshotWithNewClient`, but the JVM exits before the daemon thread completes the RPC. No `ShutdownHookManager` hook exists to drain pending cleanup. Fix: add a synchronous cleanup path in the listener (bounded timeout ~2s) plus a `ShutdownHookManager.addShutdownHook` that joins outstanding cleanup Futures.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** `planInputPartitions()` is called multiple times by Spark when AQE is enabled (default since Spark 3.2+). Each call creates a new client snapshot. Cache the result via `lazy val` or `AtomicReference` to ensure idempotency: `private lazy val plannedPartitions: Array[InputPartition] = computeInputPartitions()`.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The cleanup uses a single-threaded `ScheduledExecutorService` that also runs the blocking drop-snapshot RPC. If the RPC hangs or retries, the scheduler thread is blocked and no other scheduled tasks can fire. Split into a ScheduledExecutor for scheduling and a CachedThreadPool for RPC execution.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The 30-min `ScheduledFuture` TTL force-drops the snapshot even if a query is still running, causing mid-query data loss. It also cannot protect against JVM death (the one scenario it was designed for). Remove entirely after implementing synchronous listener cleanup + shutdown hook. Server-side `compactionProtectionSeconds` + external sweeper is the correct model.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The `compactionProtectionSeconds` option is passed to the gRPC interface without validation. Negative or zero values can cause silent corruption of compaction protection state on the server. Add strict validation rejecting values <= 0.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** `resolveClientSnapshotLocation` does not reject non-S3/S3a schemes (e.g., `gs://`). It silently prepends `s3a://`, producing an invalid path that crashes the Hadoop FileSystem layer. Throw an explicit `IllegalArgumentException` for unsupported schemes. Note: this validation must be inside the `try` block (see exception scope issue).

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** Option values parsed with `.toLong` (e.g., executionId, compactionProtectionSeconds) throw raw `NumberFormatException` on non-numeric input. Wrap in try-catch with a descriptive `IllegalArgumentException` naming the option key. Also reject negative values where semantically invalid, and apply `.filter(_.trim.nonEmpty)` to guard against empty strings.

liliu-z:
**src/test/scala/zilliztech/spark/milvus/MilvusScanClientSnapshotTest.scala:0** The test constructs `MilvusScan` with `SnapshotMode=true` and returns before any SparkSession lookup. The critical code paths — client snapshot fast path, `buildSnapshotHadoopConf`, `activeCleanupRegistration`, listener firing, retry on drop failure, AQE multi-call idempotency — are all untested. Add tests with a mocked SparkSession covering the cleanup lifecycle.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The current retry strategy for releasing snapshots totals roughly 600ms, which is insufficient to tolerate network blips or Milvus node drift. Use exponential backoff with a total window of 10-30 seconds.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** Comparing execution IDs by converting Long to String and comparing with a String is fragile and error-prone. Compare `Long` values directly to avoid subtle mismatches from formatting differences.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** `dropClientReadSnapshot` and related cleanup helpers are public on the `MilvusScan` companion object. These are internal lifecycle operations that should not be callable by external consumers. Mark as `private[sources]` or `private[connector]`.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The widened `isSnapshotMode` check should validate that the snapshot state is consistent in `initFromSnapshot()`. Add a `milvus.client.snapshot.require=true` strict-mode option so users can fail loudly rather than silently degrading to a non-snapshot scan when snapshot options are partially set.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** `MilvusScan.snapshotBucket` calls `Option(uri.getHost)`, but `new URI("s3a://my_bucket/foo").getHost` returns `null` for underscore-containing bucket names (legal in path-style S3). The bucket is silently dropped from `snapshotBucketsToConfigure`, so `fs.s3a.bucket.my_bucket.endpoint` is never configured.

liliu-z:
**src/test/scala/zilliztech/spark/milvus/MilvusScanClientSnapshotTest.scala:0** Add a reflection-based unit test that verifies all `Snapshot*` option constants in `MilvusOption` are included in the `SnapshotOptionKeys` cleanup set. This prevents future option additions from silently leaking through the cleanup filter.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** `object MilvusScan extends Logging` mixes in `private[spark]` API. Pre-existing in the codebase but compounds the version-fragility surface alongside `SparkListenerSQLExecutionEnd`. Document supported Spark version range or migrate to SLF4J directly in a separate PR.

liliu-z:
**MilvusDataSource.scala:0** `buildSnapshotHadoopConf` sets `fs.s3a.impl` unconditionally, which can overwrite a user's existing Hadoop configuration (e.g., a custom S3 filesystem implementation). Guard the assignment with `if (conf.get("fs.s3a.impl") == null)` so existing config is preserved.

liliu-z:
**MilvusOption.scala:158** `maxJsonBytes` and `compactionProtectionSeconds` use camelCase suffixes while neighboring keys use dotted lowercase (`snapshot.manifests`, `snapshot.schema.json`). Since `CaseInsensitiveStringMap` only handles case but not delimiter differences, user typos like `max-json-bytes` or `max_json_bytes` will silently fall through to defaults. Pick one convention (dotted lowercase) for this public API surface and apply it consistently.

liliu-z:
**MilvusDataSource.scala:823** When S3 credential options are absent, `buildSnapshotHadoopConf` silently falls back to the default AWS credential chain. In local-mode development, this typically fails with a 'credentials not configured' error — but only after the server-side snapshot has already been created, leaving a leaked snapshot. At minimum, log a message when using the default credential chain so users can correlate failures.

liliu-z:
**MilvusDataSource.scala:0** The `s3Bucket` resolution via `options.getOrElse(FsBucketName, ...)` appears at lines ~1034, ~706, and in `buildSnapshotHadoopConf` (~841). After removing the `"a-bucket"` default, factor this into a single helper to keep the validation logic in one place.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The `SparkListener` for snapshot cleanup is registered after the S3 JSON manifest fetch, which can take several seconds. If planning fails during that window, no listener exists to clean up the orphaned snapshot. The listener must be attached immediately after `createSnapshotForRead` succeeds, before any slow I/O. Spark's `SQLExecution.withNewExecutionId` guarantees `SQLExecutionEnd` fires in its own `finally` block, so attaching early is safe and sufficient.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** `registerClientSnapshotCleanup` uses two `AtomicBoolean` variables (`cleanupComplete` and `cleanupInFlight`) for concurrency control, but Spark's `ListenerBus` posts events serially on a single thread and `SQLExecutionEnd` fires exactly once per `executionId`. The guards add complexity without benefit and can cause the listener to never be removed if the flags get into an unexpected state. The `removeListener` call must be unconditional in a `finally` block.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** When snapshot planning fails, the catch block calls `dropClientReadSnapshot` synchronously on the Spark Driver thread. With retries, this can block the Driver for up to 25 seconds. Once the listener lifecycle is restructured (attached immediately after snapshot creation), this synchronous catch block becomes redundant — the listener handles cleanup. The only case needing a synchronous fallback is when `addSparkListener` itself fails.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The snapshot metadata deserialized from the server JSON is used without null checks. If `metadata.collection` or `metadata.collection.schema` is null, the code will throw a NullPointerException with no actionable message. The JSON is server-controlled but should not be blindly trusted.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** There is no guard asserting `MilvusOption.isSnapshotMode(snapshotOptions)` before constructing the internal `MilvusScan` in the snapshot path. Without this, a misconfigured option map could lead to infinite recursion if the code re-enters the snapshot branch unexpectedly. Adding a `require` assertion is low-cost and prevents this.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** When `isSnapshotMode` returns true (now with OR semantics) but no manifest data is present, `computeInputPartitions` calls `planInputPartitionsFromSnapshot(snapshotManifests.getOrElse(""))`. The behavior of that method with an empty string is unclear and untested. The looser predicate now reaches this code path more easily.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The snapshot name is generated by concatenating collection name with a timestamp/suffix. For long collection names, the generated snapshot name could exceed Milvus's typical 256-character limit, causing snapshot creation to fail. The failure would be caught but results in an uninformative server error.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusOption.scala:0** There are two overloads of `isSnapshotMode`: one taking `Map[String, String]` and one taking `CaseInsensitiveStringMap`. Their behavior may diverge due to case sensitivity differences, but only one overload is tested. The overloaded method tests should be consistent and both should be covered.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The `MilvusScan` companion object initializes `CleanupRpcExecutor` (a thread pool) and registers a JVM shutdown hook on first reference, which happens on every executor JVM. However, cleanup only ever runs on the driver. This wastes one thread and one shutdown hook per executor.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** The cleanup path passes the entire captured user options map to `MilvusClient(MilvusOption(baseOptions))`. Currently `MilvusClient` only reads connection options from this map, but if it ever reacts to `SnapshotMode` or other options, the cleanup client could misbehave. Minor coupling risk.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusScan.scala:0** When snapshot metadata references S3 paths, the code uses the configured credentials. If the snapshot data resides in a different bucket than the connector's configured bucket, the credentials may not have access. No warning or validation is performed.

liliu-z:
**MilvusSnapshotReader.scala:0** The 64 MB size limit (`SnapshotMaxJsonBytes`) is now also enforced on local `Source.fromFile` reads. Users who previously loaded snapshot JSON files larger than 64 MB from local disk will now get an error. This is a reasonable safety change but is a backwards-incompatible behavioral change that must be documented in Release Notes.

liliu-z:
**MilvusOptionTest.scala:0** There is no negative test that verifies the fix for explicit `SnapshotMode=false` taking precedence over implicit conditions like manifests being present. Without this test, the semantic fix for the explicit-false-wins behavior has no regression guard.

liliu-z:
**MilvusSnapshotReader.scala:0** When both `manifestList` and `storageV2ManifestList` are empty in a client snapshot, the code proceeds to perform S3 reads and JSON parsing before eventually hitting `IllegalArgumentException` in the inner planner's explicit-mode check. A pre-flight check in `planInputPartitionsFromClientSnapshotPath` would fail faster with a specific message ('client snapshot is empty: no manifests and no V2 segments').

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** When `fs.s3a.impl.disable.cache=true` is set (as recommended to prevent credential pollution), `FileSystem.get` creates a new `S3AFileSystem` instance on every call. The current code only closes the `InputStream` but never calls `fs.close()`. Each invocation permanently leaks an S3A HTTP connection pool, retry threads, and statistics objects in the Driver JVM, leading to thread exhaustion or OOM after a few queries. Wrap the entire block in a `try/finally` that calls `fs.close()` after the inner `in.close()`.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** Multiple concurrent Spark jobs sharing the same JVM can pollute each other's Hadoop S3A credentials via the shared `FileSystem` cache. `buildSnapshotHadoopConf` must inject `fs.s3a.impl.disable.cache=true` to force per-call `FileSystem` instantiation and prevent cross-job credential leakage.

liliu-z:
**src/main/scala/sources/MilvusDataSource.scala:541** `pushFilters` at line 541 checks `options.get(SnapshotV2Segments)` before a client snapshot exists, so it classifies filters as supported. Later, the client-snapshot path discovers V2 segments and recurses, but the V2 reader never applies those pushed filters. Spark sees `pushedFilters()` return non-empty and assumes the source handled them — it won't re-apply them. This silently returns unfiltered rows.

Fix: Disable the client-snapshot fast path entirely when `pushedFilterArray` is non-empty, or return all filters as unsupported for client-snapshot scans until V2 filtering is implemented.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -647,52 +1031,228 @@ class MilvusScan(
...
+      }
+    } finally {
+      client.close()
+    }
+  }
+
+  private def planInputPartitionsFromClientSnapshot(
+      client: MilvusClient
+  ): Option[Array[InputPartition]] = {
+    val s3Bucket = MilvusScan.resolveConnectorS3Bucket(milvusOption.options)
```
liliu-z: `resolveConnectorS3Bucket` at line 1099 throws synchronously before the snapshot RPC is attempted. On older Milvus servers that would return `UNIMPLEMENTED`, this now fails early instead of falling back to `GetPersistentSegmentInfo`. The legacy path doesn't need the bucket name — it builds paths from `FsRootPath`. This is a backwards-incompatible behavior change for users on legacy Milvus who never configured a bucket.

Fix: Move bucket resolution after snapshot RPC success, or treat missing bucket as "skip fast path, log warning".

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -757,29 +1312,98 @@ class MilvusScan(
...
-    logInfo(
-      s"Created ${partitions.length} partitions (treating all StorageV2/V3 " +
-        "segments as V3; see V2 dispatch TODO)"
-    )
-    client.close()
+    logInfo(s"Created ${partitions.length} partitions via legacy client path")
     partitions
   }
 
+  private[sources] def buildSnapshotHadoopConf(
```
liliu-z, line 1527: `buildSnapshotHadoopConf` maps endpoint/AK/SK/SSL but not `fs.region`, `fs.use_iam`, or `fs.use_virtual_host`. It only honors raw `fs.s3a.path.style.access`. MinIO setups using `fs.use_virtual_host=false` will hit path-style errors. Non-default AWS regions will fail to resolve. The legacy read path handles these correctly, so the snapshot path silently regresses for MinIO/IRSA/non-us-west-2 deployments.

Fix: Add a helper that maps connector keys to S3A keys, sharing the truth source with `Properties.scala`.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -59,9 +77,7 @@ case class MilvusDataSource() extends TableProvider with DataSourceRegister {
   ): Table = {
     val options = new CaseInsensitiveStringMap(properties)
     val milvusOption = MilvusOption(options)
-    val isSnapshotMode =
-      Option(options.get(MilvusOption.SnapshotMode)).contains("true") ||
-        Option(options.get(MilvusOption.SnapshotManifests)).isDefined
+    val isSnapshotMode = MilvusOption.isSnapshotMode(options)
     if (milvusOption.uri.isEmpty && !isSnapshotMode) {
```
liliu-z, line 83: `createReaderFactory()` at line 1554 is called on the outer `MilvusScan`. The outer scan's `options` lacks `SnapshotMode`, `SnapshotManifests`, `SnapshotV2Segments` — those only exist on the inner recursive scan. If `MilvusPartitionReaderFactory` inspects `MilvusOption.isSnapshotMode(options)` at task time to choose a code path, it will see the outer options and make the wrong decision. Needs end-to-end tracing to confirm whether partition-level state is sufficient.

liliu-z:
**src/main/scala/sources/MilvusParquetFooterReader.scala:0** The new helper uses `new URI(path)` which will throw on S3 object keys containing literal spaces, `#`, or other reserved characters. The previous code used `HadoopInputFile.fromPath(new Path(path), ...)` which handled this via `Path`'s more lenient parsing. This is a regression for any object keys with unescaped reserved characters.

Fix: Use `new Path(path).toUri` instead of `new URI(path)`, or document the escaping requirement.

liliu-z:
**src/main/scala/sources/MilvusDataSource.scala:378** The `addShutdownHook` at line 378 runs when `MilvusScan` is class-loaded. If the class is loaded just for inspection (e.g., reflection in a test harness), a shutdown hook is registered even if no snapshot is ever created.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +633,406 @@ class MilvusScanBuilder(
...
+  private def daemonThreadFactory(namePrefix: String): ThreadFactory =
+    new ThreadFactory {
+      override def newThread(r: Runnable): Thread = {
+        val thread = new Thread(r, namePrefix)
+        thread.setDaemon(true)
+        thread
+      }
+    }
+
+  private val CleanupRpcExecutor = Executors.newCachedThreadPool(
```
liliu-z: Uses `Executors.newCachedThreadPool` with no upper bound. The retry loops at lines 858-892 include `Thread.sleep` calls that block threads for up to ~13s per attempt. Under load (many concurrent queries with snapshot cleanup), this can spawn an unbounded number of threads and exhaust system resources. Replace with a small bounded executor (e.g. `newFixedThreadPool` or `newScheduledThreadPool`).

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -163,11 +177,8 @@ case class MilvusTable(
     * [[MilvusOption.SnapshotV2Segments]] is sufficient to take the
     * snapshot-only path and skip all Milvus client calls.
     */
-  private def isSnapshotMode: Boolean = {
-    milvusOption.options.get(MilvusOption.SnapshotMode).contains("true") &&
-    (milvusOption.options.contains(MilvusOption.SnapshotManifests) ||
-      milvusOption.options.contains(MilvusOption.SnapshotV2Segments))
-  }
+  private def isSnapshotMode: Boolean =
```
liliu-z, line 184: `isSnapshotMode` now returns true if *any* snapshot-related option is set, not just when `SnapshotMode=true`. This means setting `SnapshotMode=true` alone now silently bypasses the client path where previously it would fall through. This is a compatibility-breaking semantic change that should either be validated earlier in `MilvusTable`/`inferSchema` or explicitly documented as an intentional change.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +633,406 @@ class MilvusScanBuilder(
...
+      options: scala.collection.Map[String, String]
+  ): String = {
+    connectorS3BucketOption(options).getOrElse {
+      throw new IllegalArgumentException(
+        s"${Properties.FsConfig.FsBucketName} is required for client snapshot reads"
+      )
+    }
+  }
+
+  private[sources] def snapshotS3BucketForRelativePaths(
```
liliu-z, line 799: Line 775 prefers the connector bucket over the bucket from the resolved snapshot path when computing `snapshotS3BucketForRelativePaths`. This propagates to `V2SegmentLoader` at line 1260 and through `buildClientSnapshotOptions` at line 981 for V3. In cross-cluster snapshot-share scenarios (snapshot metadata in bucket A, connector pointed at bucket B), relative paths inside the snapshot metadata will resolve against the wrong bucket. The correct precedence depends on Milvus's path convention (relative to storage root vs. relative to snapshot s3_location). This needs explicit clarification from the author.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -647,52 +1058,257 @@ class MilvusScan(
...
         )
+        if (
+          connectorBucket.isEmpty && MilvusScan
+            .isBucketRelativeSnapshotLocation(snapshot.s3Location)
+        ) {
+          logWarning(
+            s"Skipping client snapshot fast path because ${Properties.FsConfig.FsBucketName} is missing " +
+              "and the snapshot location is bucket-relative; falling back to legacy GetPersistentSegmentInfo read path"
+          )
+          MilvusScan
```
liliu-z: Fallback planning at lines 1168-1182 and 1202-1217 calls `dropClientReadSnapshot` synchronously on the driver-side planning thread. With retry and sleep, this can block for ~13 seconds, stalling the Spark planner. Queue cleanup asynchronously or use a single short-timeout attempt during planning.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +633,406 @@ class MilvusScanBuilder(
...
+      case None if bucket.trim.nonEmpty =>
+        s"s3a://${bucket.trim}/${trimmed.stripPrefix("/")}"
+      case None =>
+        throw new IllegalArgumentException(
+          "bucket-relative snapshot s3_location requires connector S3 bucket"
+        )
+    }
+  }
+
+  private[sources] def snapshotBucket(location: String): Option[String] = {
```
liliu-z, line 740: Lines 724-737 call `resolveClientSnapshotLocation(trimmed, "")` with an empty bucket. If the input is bucket-relative (no scheme), this throws `IllegalArgumentException`. Current call sites always pass resolved paths, but the helper is a footgun — it should either return `None` for bucket-relative inputs or be renamed to clarify it requires a fully-qualified URI.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +633,406 @@ class MilvusScanBuilder(
...
+
+  private[sources] def canUseClientSnapshotFastPath(
+      milvusOption: MilvusOption
+  ): Boolean = {
+    milvusOption.partitionName.isEmpty &&
+    milvusOption.partitionID.isEmpty &&
+    milvusOption.segmentID.isEmpty
+  }
+
+  private[sources] def generatedClientSnapshotName(
```
liliu-z, line 870: Lines 797-808 use the collection name directly in the generated snapshot name. Milvus snapshot names typically allow only alnum + `_`. A collection name containing `-`, `.`, or unicode characters will produce an invalid snapshot name and `createSnapshot` will fail. Sanitize the collection name (e.g. regex-replace non-alnum to `_`) or document that `ClientSnapshotName` must be set manually for such collections.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +633,406 @@ class MilvusScanBuilder(
...
+  private[sources] def buildClientSnapshotOptions(
+      baseOptions: Map[String, String],
+      collectionName: String,
+      collectionId: Long,
+      partitionIds: Seq[Long],
+      schemaBytesBase64: String,
+      manifestList: Seq[StorageV2ManifestItem],
+      v2Segments: Seq[V2SegmentInfo]
+  ): Map[String, String] = {
+    var out = baseOptions.filterNot { case (key, _) =>
```
liliu-z, line 1129: In the V3 recursive path, `buildClientSnapshotOptions` preserves the original `fs.bucket_name`, so relative manifest/base paths in the FFI reader still use the old connector bucket rather than the snapshot bucket. This compounds the cross-bucket issue from line 775 for V3 segment reading.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -45,7 +58,12 @@ import com.zilliz.spark.connector.loon.Properties
 import com.zilliz.spark.connector.read.{
   MilvusPackedV2InputPartition,
   MilvusPartitionReaderFactory,
-  MilvusStorageV3InputPartition
+  MilvusSnapshotReader,
+  MilvusStorageV3InputPartition,
+  SnapshotMetadata,
+  StorageV2ManifestItem,
```
liliu-z, line 65: Several locations use `s"fs.${uri.getScheme}.impl.disable.cache"` without guarding against `uri.getScheme` returning null (which it does for scheme-less URIs). This produces a key like `fs.null.impl.disable.cache` which is harmless but semantically wrong. Guard with a null check or validate the URI has a scheme.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The `plannedPartitions` lazy val caches partition data for the lifetime of the `MilvusScan` instance. When a client-snapshot is cleaned up after the first action, repeated DataFrame actions (e.g., `.count()` then `.show()`) silently reuse partitions pointing to a dropped snapshot, producing empty or corrupt results. This is a "silently wrong" failure — no error is raised.

Fix: Remove caching for the client-snapshot path, or scope it to the SQL execution lifetime.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** When connector-supplied AK/SK credentials are used (non-IAM branch), `buildSnapshotHadoopConf` does not explicitly set `fs.s3a.aws.credentials.provider` to `SimpleAWSCredentialsProvider`. The inherited session Hadoop config may contain an IRSA/EKS credential provider chain that silently overrides the connector-supplied credentials. This causes silent authentication failures or wrong-account access in IRSA/EKS deployments.

Fix: Explicitly set `SimpleAWSCredentialsProvider` (or equivalent) in the non-IAM branch, both globally and per-bucket.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusParquetFooterReader.scala:157** In `readWithFileSystem`, the `URI` is constructed before the `try` block. If the path is malformed, the resulting exception escapes uncaught, violating the declared `Either[Throwable, T]` return contract of `read` and `readFieldIdsFromSchema`. Callers relying on `Left` for error handling will get an unexpected throw instead.

Fix: Move URI construction inside the `try` block. Also add `Option(uri.getScheme)` defense in the close/cleanup path.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** The relaxed `isSnapshotMode` semantics — where merely defining `SnapshotManifests` without explicitly setting `SnapshotMode=true` enters snapshot mode — changes behavior in `getTable`, `inferSchema`, `MilvusTable.initInfo`, and `MilvusScan.computeInputPartitions`. This behavioral change is not mentioned in the PR description or release notes. If intentional, it needs tests covering all call sites and documentation. If unintentional, require explicit `SnapshotMode=true`.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/V2SegmentLoader.scala:0** Similar to the `MilvusParquetFooterReader` issue, `V2SegmentLoader.readAllBytes` constructs a URI outside error handling. Since this method is throw-based rather than `Either`-based, the impact is lower — the error message is just less helpful rather than breaking a contract.

Fix: Move URI construction inside try for consistency and better error messages.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:669** A `Runtime.getRuntime.addShutdownHook` is registered at companion object initialization time — the first time anything touches `MilvusScan`. This means the hook is installed even if no client snapshot is ever created. In multi-classloader scenarios (Spark Connect, structured streaming reloads), the second registration with the same thread name throws `IllegalArgumentException` and breaks class loading.

Fix: Wrap the hook registration in a `Try` to tolerate multi-classloader edge cases.

liliu-z:
**src/main/scala/zilliztech/spark/milvus/MilvusDataSource.scala:0** If `SparkListenerSQLExecutionEnd` never fires (e.g., Spark Connect mode, listener registration race), the registered cleanup callback is never invoked and the client-read snapshot leaks. The shutdown hook provides a last-resort drain, but a fallback timer would catch the intermediate case.

Fix: Add a fallback cleanup timer for the registered-but-listener-misses-event race.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +635,453 @@ class MilvusScanBuilder(
...
+  }
+
+  private def dropClientReadSnapshotWithNewClient(
+      baseOptions: Map[String, String],
+      databaseName: String,
+      collectionName: String,
+      snapshotName: String,
+      reason: String
+  ): Try[Unit] = {
+    Try(MilvusClient(MilvusOption(baseOptions))).flatMap { client =>
```
liliu-z, line 1023: `client.close()` after a successful `dropClientReadSnapshot` call can throw and shadow the successful drop result. If `close()` throws, the caller sees a failure even though the snapshot was dropped successfully. Wrap `close()` in its own `Try`, log any close failure, and preserve the drop result.

liliu-z:
**src/main/scala/loon/Properties.scala:17** The native FFI reader uses a single `fs.bucket_name` (set here), but `MilvusDataSource.scala:1040` can replace the bucket with the snapshot JSON bucket. Hadoop per-bucket config only helps the driver read JSON/AVRO; executors still read parquet through one native bucket. If snapshot metadata and data parquet reside in different buckets, reads will silently target the wrong bucket. Add validation or fail-fast when `snapshotBucket != connectorBucket` and paths are bucket-relative.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1,11 +1,21 @@
 package com.zilliz.spark.connector.sources
 
 import java.{util => ju}
-import java.util.{HashMap, Map => JMap}
+import java.net.URI
+import java.util.{Base64, HashMap, Map => JMap, UUID}
+import java.util.concurrent.{ConcurrentHashMap, Executors, ThreadFactory}
+import java.util.concurrent.atomic.AtomicBoolean
 import scala.collection.mutable
+import scala.concurrent.{Await, ExecutionContext, Future}
```
liliu-z, line 9: The shutdown hook drains pending cleanup futures, but there is a race: new futures can be added after the drain snapshot is taken. Use `CompletableFuture.allOf(...)` on a snapshot of pending futures plus a second pass, or otherwise ensure no new futures are submitted after the drain begins.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -77,9 +94,8 @@ case class MilvusDataSource() extends TableProvider with DataSourceRegister {
     val milvusOption = MilvusOption(options)
 
     // Check for snapshot mode - use snapshot schema if provided
-    val isSnapshotMode =
-      Option(options.get(MilvusOption.SnapshotMode)).contains("true") ||
-        Option(options.get(MilvusOption.SnapshotManifests)).isDefined
+    MilvusOption.validateSnapshotModeOptions(options)
```
liliu-z, line 98: The `milvus.snapshot.max.json.bytes` option is parsed without reusing `parsePositiveLongOption`, leading to inconsistent error handling for malformed values. Reuse the existing helper.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -59,9 +77,8 @@ case class MilvusDataSource() extends TableProvider with DataSourceRegister {
   ): Table = {
     val options = new CaseInsensitiveStringMap(properties)
     val milvusOption = MilvusOption(options)
-    val isSnapshotMode =
-      Option(options.get(MilvusOption.SnapshotMode)).contains("true") ||
-        Option(options.get(MilvusOption.SnapshotManifests)).isDefined
+    MilvusOption.validateSnapshotModeOptions(options)
```
liliu-z, line 81: There is no upper bound or warning on `compactionProtectionSeconds`. A user could set an extremely large value, causing snapshots to be held indefinitely and blocking compaction on the Milvus server. Add an upper bound check with a warning.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -45,7 +58,12 @@ import com.zilliz.spark.connector.loon.Properties
 import com.zilliz.spark.connector.read.{
   MilvusPackedV2InputPartition,
   MilvusPartitionReaderFactory,
-  MilvusStorageV3InputPartition
+  MilvusSnapshotReader,
+  MilvusStorageV3InputPartition,
+  SnapshotMetadata,
+  StorageV2ManifestItem,
+  V2SegmentInfo,
+  V2SegmentLoader
```
liliu-z, line 67: New `catch Throwable` blocks in `V2SegmentLoader.readAllBytes` and `MilvusParquetFooterReader.readWithFileSystem` catch fatal errors like `OutOfMemoryError` and `StackOverflowError`. Use `NonFatal` to avoid swallowing fatal exceptions.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -24,6 +34,8 @@ import org.apache.spark.sql.connector.read.{
   SupportsPushDownRequiredColumns
 }
 import org.apache.spark.sql.connector.write.{LogicalWriteInfo, WriteBuilder}
+import org.apache.spark.sql.execution.ui.SparkListenerSQLExecutionEnd
```
liliu-z, line 37: The code uses `SparkListenerSQLExecutionEnd` and `SQLExecution.EXECUTION_ID_KEY` which are Spark internal APIs not covered by Spark's stability guarantees. These may break across Spark versions. Add a code comment documenting the Spark version constraint and the risk.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -428,12 +457,11 @@ object MilvusSnapshotReader {
       path: String
   ): Either[Throwable, SnapshotMetadata] = {
     try {
-      val source = scala.io.Source.fromFile(path)
+      val in = new java.io.FileInputStream(path)
       try {
```
liliu-z, line 576: `toProtobufSchemaBytes` preserves `fieldID`, `name`, `dataType`, `isPrimaryKey`, `isClusteringKey`, and `typeParams` (including vector `dim`), but silently drops: schema-level `enableDynamicField`, `autoID`, `properties`; per-field `defaultValue`, `nullable`, `state`, `isFunctionOutput`, `elementType` (arrays), `isPartitionKey`. If any of these are needed on the read path (especially `isPartitionKey` for partition pruning and `elementType` for array typing), the round-trip is lossy. Add a round-trip test asserting these specific fields are preserved or documented as out-of-scope.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -647,52 +1107,234 @@ class MilvusScan(
...
+      milvusOption.options
+    )
+    val v2Segments =
+      if (metadata.manifestList.nonEmpty) {
+        V2SegmentLoader.loadV2Segments(
+          metadata.manifestList,
+          snapshotBucket.getOrElse(""),
+          hadoopConf
+        ) match {
+          case Right(segs) => segs
```
liliu-z, line 1392: When `metadata.manifestList` contains only entries with `storage_version != 2`, `V2SegmentLoader.loadV2Segments` skips all of them and returns `Right(Seq.empty)`. Combined with an empty `metadata.storageV2ManifestList`, the resulting snapshot options carry no segment data and the planner silently emits zero partitions. The legacy client path throws `No packed-parquet segments (StorageV2/V3) found` for the same situation; the upfront `validateClientSnapshotMetadata` check only rejects when both manifest lists are empty before parsing, not after non-V2 entries are filtered out.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -586,20 +727,17 @@ object MilvusSnapshotReader {
...
+      case 24  => StringType // Geometry
+      case 25  => StringType // Text
+      case 26  => LongType // Timestamptz
+      case 100 => ArrayType(ByteType) // BinaryVector
       case 101 => ArrayType(FloatType) // FloatVector
-      case 102 => ArrayType(ByteType) // BinaryVector
-      case 103 => ArrayType(ShortType) // Float16Vector
-      case 104 => ArrayType(ShortType) // BFloat16Vector
-      case 105 => MapType(LongType, FloatType) // SparseFloatVector
+      case 102 => ArrayType(FloatType) // Float16Vector
```
liliu-z, line 742: The snapshot schema now maps `BinaryVector`/`Int8Vector` to `ArrayType(ByteType)` and `Float16Vector`/`BFloat16Vector` to `ArrayType(FloatType)`, but `ArrowConverter` only handles the 4-byte `FloatVector` case for `ArrayType(FloatType)`: `val floats = (0 until (bytes.length / 4)).map(_ => buffer.getFloat())` (ArrowConverter.scala:96-102) reads 4-byte floats, so a 128-dim Float16Vector (256 bytes) yields 64 garbage floats with bits taken from adjacent half-floats. `BinaryVector`/`Int8Vector` fall into the generic array path that casts a `FixedSizeBinaryVector` to `ListVector`, which throws `ClassCastException` at task time. A snapshot containing any of these vector types either silently corrupts data (Float16/BFloat16) or fails materialization (Binary/Int8); fix by adding width-aware decoders or by keeping these columns as raw binary until `ArrowConverter` is extended (compare `FloatConverter.fromFloat16Bytes` at `MilvusUtil.scala:436`).

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +615,566 @@ class MilvusScanBuilder(
...
+        case Failure(e) =>
+          lastFailure = Some(e)
+          logWarning(
+            s"Failed to drop client read snapshot $snapshotName after $reason " +
+              s"(attempt $attempt/$maxAttempts)",
+            e
+          )
+          if (attempt < maxAttempts) {
+            val delayMillis = InitialDropRetryDelayMillis << (attempt - 1)
+            Thread.sleep(delayMillis)
```
liliu-z, line 992: The retry loop `case Failure(e) => ... Thread.sleep(InitialDropRetryDelayMillis << (attempt - 1))` gates only on attempt count, so a failed drop accumulates 200+400+800+1600+3200 ≈ 7.2 s of sleep (12.6 s if the final attempt sleeps before exiting) regardless of the underlying RPC code; `MilvusClient.dropSnapshot` (MilvusClient.scala:739) returns a raw `Failure(e)` for both transient and terminal errors, and only `isServiceNotImplemented` is special-cased — there is no NOT_FOUND/PERMISSION_DENIED fast-fail. A snapshot already GC'd by the server or a collection with an ACL change will burn the full backoff per call, blocking one of the two cleanup workers; concurrent failures then queue behind the sleeps and the 2 s shutdown drain timeout leaves snapshots orphaned. Short-circuit terminal status codes (NOT_FOUND, PERMISSION_DENIED, INVALID_ARGUMENT) before sleeping, similar to the existing `isServiceNotImplemented` branch.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -622,6 +615,566 @@ class MilvusScanBuilder(
...
+    }
+  }
+
+  private[sources] def validateSnapshotBucketForRelativeDataPaths(
+      snapshotPath: String,
+      connectorBucket: Option[String],
+      storageV2ManifestList: Seq[StorageV2ManifestItem],
+      v2Segments: Seq[V2SegmentInfo]
+  ): Unit = {
+    (snapshotBucket(snapshotPath), connectorBucket) match {
```
liliu-z: The guard `(snapshotBucket(snapshotPath), connectorBucket) match { case (Some(snapshot), Some(connector)) if snapshot != connector =>` only fires when both buckets are `Some`, but the legacy-path fallback at line 1291 only kicks in when the snapshot location itself is bucket-relative. The case `connectorBucket=None` + a fully qualified `s3a://snapshot-bucket/...` snapshot path + a manifest whose `base_path` is bucket-relative slips past both checks, and executors silently target the snapshot bucket. If actual data lives in a different bucket, every task fails with file-not-found at runtime instead of surfacing the driver-side validation message this PR introduced. Trigger the cross-bucket check whenever `snapshotBucket(snapshotPath)` is `Some`, even when the connector bucket is unset.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -93,13 +110,33 @@ object ArrowConverter extends Logging {
...
+        ArrayData.toArrayData(bytes)
+
+      case ArrayType(ShortType, _)
+          if vector.isInstanceOf[FixedSizeBinaryVector] =>
+        val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
+        ArrayData.toArrayData(bytes.map(_.toShort))
+
+      case ArrayType(FloatType, _)
+          if vector.isInstanceOf[FixedSizeBinaryVector] =>
+        val bytes = vector.asInstanceOf[FixedSizeBinaryVector].get(rowIndex)
```
liliu-z, line 132: `MilvusDataSource.supportsExternalMetadata = true` lets callers pass their own `StructType` via `.schema(...)`, and `MilvusTable.schema()` returns that user schema verbatim in snapshot mode. User-supplied `StructField`s won't carry the `milvus.data_type` metadata that the new FP16/BF16 decoding relies on. For a Float16Vector/BFloat16Vector column the Arrow data is a 2-byte FixedSizeBinaryVector but the user marks it `ArrayType(FloatType)`; `arrowValueToSparkValue` then calls `decodeFixedSizeBinaryFloats(bytes, None)`, which falls into the `_ =>` branch and reads the bytes as 4-byte little-endian floats, silently corrupting every vector. Either reject user schemas missing the milvus type metadata for vector columns, or pick the decoder from FixedSizeBinary element width rather than from optional metadata.

PR #97, opened 2026-05-29
enhance!: [reader] canonicalize metadata extra columns

congqixia:
Freeze metadata extra-column output names as `partition`, `$segment_id`, and `$row_offset`, while preserving `segment_id` / `row_offset` as input aliases. Update reader metadata wrapping and backfill consumers to use canonical names without changing snapshot lifecycle or snapshot option keys.

BREAKING CHANGE: `milvus.extra.columns=segment_id,row_offset` still works as input, but emitted schemas now use `$segment_id` and `$row_offset`.

### DIFF
--- a/src/main/scala/read/MilvusPartitionReaderFactory.scala
+++ b/src/main/scala/read/MilvusPartitionReaderFactory.scala
@@ -10,6 +10,7 @@ import org.apache.spark.sql.connector.read.{
 import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types.StructType
 
+import com.zilliz.spark.connector.MilvusOption
 import io.milvus.grpc.schema.CollectionSchema
 
 // PartitionReaderFactory for Storage V2 (Milvus 2.6+)
@@ -20,6 +21,14 @@ class MilvusPartitionReaderFactory(
 ) extends PartitionReaderFactory
     with Logging {
 
+  private def isSystemField(name: String): Boolean =
+    name == "row_id" || name == "timestamp"
+
+  private def isMetadataExtraField(name: String): Boolean =
+    name == MilvusOption.MilvusExtraColumnPartition ||
+      name == MilvusOption.MilvusExtraColumnSegmentID ||
+      name == MilvusOption.MilvusExtraColumnRowOffset
+
   override def createReader(
       partition: InputPartition
   ): PartitionReader[InternalRow] = {
@@ -29,13 +38,8 @@ class MilvusPartitionReaderFactory(
           s"Creating V3 reader for partition with segmentID=${p.segmentID}"
         )
 
-        // Storage V3 doesn't support system fields (row_id, timestamp) and extra metadata columns (segment_id, row_offset)
-        // Filter them out from the schema for the underlying reader
-        val v2Schema = StructType(schema.fields.filter { field =>
-          field.name != "row_id" &&
-          field.name != "timestamp" &&
-          field.name != "segment_id" &&
-          field.name != "row_offset"
+        val v2Schema = StructType(schema.fields.filterNot { field =>
+          isSystemField(field.name) || isMetadataExtraField(field.name)
         })
 
         // Deserialize the protobuf schema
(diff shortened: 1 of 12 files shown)


src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,55 +60,34 @@ class MilvusPartitionReaderFactory(
...
-              }
-              if (hasRowOffset) {
-                resultValues(writeIdx) = rowOffset
-                rowOffset += 1
-                writeIdx += 1
+              schema.fields.zipWithIndex.foreach { case (field, writeIdx) =>
+                field.name match {
+                  case "row_id" | "timestamp" =>
+                    resultValues(writeIdx) = null
+                  case MilvusOption.MilvusExtraColumnPartition =>
```
liliu-z, line 113: `isMetadataExtraField` classifies any field literally named `partition` (and `$segment_id`/`$row_offset`) as metadata, with no reference to whether the read actually requested those extras. The conflict guard in `appendExtraColumns` only fires when both the user schema has `partition` AND `milvus.extra.columns` contains `partition`, so a Milvus collection that simply has a user field called `partition` (without requesting it as metadata) silently passes through `MilvusTable.schema()` — but then in the reader factory, `hasSyntheticFields` flips true on the name match, the user's field is stripped from the underlying reader's projection, and the wrapper overwrites the slot with `p.partitionName`. Drive metadata classification from the requested extra-column set (`milvusOption.extraColumns`) instead of the literal field name so the reserved-name collision can only fire when the caller explicitly asked for the metadata column.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,55 +60,34 @@ class MilvusPartitionReaderFactory(
...
+                    resultValues(writeIdx) = p.partitionName
```
liliu-z: Both the V3 wrapper (`resultValues(writeIdx) = p.partitionName` at line 81) and the V2-packed wrapper (`out(writeIdx) = p.partitionID.toString` at line 142) write a raw `java.lang.String` into a slot whose `StructField` is `StringType`. Spark's `InternalRow` representation requires `UTF8String` for `StringType` (see `ArrowConverter.scala:118` `UTF8String.fromBytes(bytes)` for the same field type), so the returned row will throw `ClassCastException` the first time any downstream operator calls `getUTF8String` / converts to `UnsafeRow`. This fires for every read that requests `partition` via `milvus.extra.columns`, which is the documented use case. Wrap both values in `UTF8String.fromString(...)` before assignment.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -328,14 +328,44 @@ case class MilvusTable(
...
+
   override def schema(): StructType = {
     // In snapshot mode with provided sparkSchema, use it directly
     // This avoids the need to parse milvusCollection.schema which may be incomplete
     if (isSnapshotMode && sparkSchema.isDefined && sparkSchema.get.nonEmpty) {
       logInfo(
         s"Using provided sparkSchema in snapshot mode: ${sparkSchema.get.fieldNames.mkString(", ")}"
       )
-      return sparkSchema.get
+      return appendExtraColumns(sparkSchema.get)
```
liliu-z: In snapshot mode the user-supplied schema used to be returned verbatim; this PR now feeds it through `appendExtraColumns`, which only fails-fast on a name conflict against the canonical names (`$segment_id`, `$row_offset`, `partition`). A pre-existing snapshot caller whose schema still uses the legacy alias `segment_id`/`row_offset` falls through the conflict check (the names differ), gets the canonical column appended alongside it, and the user's legacy-named field becomes a regular data column the reader will fail to populate (or fill with null). The new test `table schema keeps legacy alias user field separate from canonical metadata` documents exactly this outcome (`schema.fieldNames == Seq("segment_id", "$segment_id")`), and the BREAKING CHANGE note only mentions emitted output names, not this silent input-schema fallback. Either normalize legacy metadata names in caller-provided schemas, or fail-fast when a provided schema contains the legacy aliases.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -328,14 +328,73 @@ case class MilvusTable(
...
+        fields = fields :+ field
+      }
+    }
+
+    if (
+      milvusOption.extraColumns.contains(
+        MilvusOption.MilvusExtraColumnSegmentID
+      )
+    ) {
+      failIfPresent(
```
liliu-z, line 361: At line 358, `failIfPresent(MilvusOption.MilvusExtraColumnSegmentIDAlias, MilvusOption.MilvusExtraColumnSegmentID)` rejects any base schema containing a user field literally named `segment_id`, which defeats the PR's stated purpose — the `$`-prefix exists precisely so a user `segment_id` column can coexist with metadata `$segment_id`. In client-derived schema mode, `baseSchema` carries the real user field; `milvus.extra.columns=segment_id` normalizes to `$segment_id`; and this guard throws before the read starts, even though the output should contain both. Restrict the legacy-alias rejection to user-provided snapshot schemas (where it would catch old manually-added metadata columns) instead of all base schemas.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -56,55 +93,35 @@ class MilvusPartitionReaderFactory(
...
-              }
-              if (hasRowOffset) {
-                resultValues(writeIdx) = rowOffset
-                rowOffset += 1
-                writeIdx += 1
+              schema.fields.zipWithIndex.foreach { case (field, writeIdx) =>
+                field.name match {
+                  case "row_id" | "timestamp" =>
+                    resultValues(writeIdx) = null
+                  case MilvusOption.MilvusExtraColumnPartition =>
```
liliu-z, line 113: At line 113, `resultValues(writeIdx) = stringValue(p.partitionName)` reports the wrong partition for V3 snapshot reads because snapshot planning stores `defaultPartitionId = partitionIds.headOption.getOrElse("0")` (MilvusDataSource.scala:1721) as `p.partitionName` for every V3 manifest item. On a multi-partition snapshot with `milvus.extra.columns=partition`, rows from later partitions all silently report the first partition's id. The wrapper should derive the partition id/name from the manifest path or per-segment metadata, not the shared default placeholder.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1709,28 +1751,32 @@ class MilvusScan(
...
+        } catch {
+          case _: NumberFormatException => 0L
+        }
+      } else 0L
+      val insertLogIdx = pathParts.lastIndexOf("insert_log")
+      val partitionId =
+        if (insertLogIdx >= 0 && pathParts.length > insertLogIdx + 3) {
+          pathParts(insertLogIdx + 2)
+        } else {
+          defaultPartitionId
```
liliu-z, line 1774: When the manifest basePath lacks the expected `insert_log/{collectionID}/{partitionID}/{segmentID}` layout, `partitionId` falls back to `defaultPartitionId` (the first requested partition id, else "0"). The following `logInfo` reports the chosen id but does not distinguish a parsed id from a fallback, so a malformed or unexpected manifest layout silently mislabels the `partition` metadata column for every affected row. Log a warning when the fallback path is taken.

PR #98, opened 2026-06-02
enhance: [reader] unify system field aliases around canonical RowID/Timestamp

congqixia:
The reader stack still exposed lowercase system-field names on some paths while snapshot and packed-V2 code already treated RowID and Timestamp as canonical. That split made projection behavior inconsistent and forced the reader factory to synthesize null system columns instead of reading the real values.

Standardize on RowID and Timestamp as the public contract, while keeping row_id, rowid, and timestamp as compatible read-time aliases. This aligns schema inference with the underlying storage field IDs and avoids collisions with Milvus-reserved system-field slots.

### DIFF
--- a/src/main/scala/read/MilvusPackedV2PartitionReader.scala
+++ b/src/main/scala/read/MilvusPackedV2PartitionReader.scala
@@ -25,6 +25,8 @@ import io.milvus.storage.{
 }
 
 object MilvusPackedV2PartitionReader {
+  private val ToleratedUnmappedColumns = Set("$meta")
+
   private[read] case class FieldMappings(
       fieldIdToName: Map[Long, String],
       fieldNameToId: Map[String, Long],
@@ -89,8 +91,9 @@ object MilvusPackedV2PartitionReader {
         }
         neededColumnFieldIds
       } else {
-        val missingNames = sourceSchema.fieldNames.filterNot(
-          fieldMappings.fieldNameToId.contains
+        val missingNames = sourceSchema.fieldNames.filterNot(name =>
+          fieldMappings.fieldNameToId.contains(name) ||
+            ToleratedUnmappedColumns.contains(name)
         )
         if (missingNames.nonEmpty) {
           throw new IllegalArgumentException(
@@ -98,7 +101,7 @@ object MilvusPackedV2PartitionReader {
               s"; schema columns=${fieldMappings.fieldNameToId.keys.toSeq.sorted.mkString(",")}"
           )
         }
-        sourceSchema.fieldNames.toSeq.map(fieldMappings.fieldNameToId)
+        sourceSchema.fieldNames.toSeq.flatMap(fieldMappings.fieldNameToId.get)
       }
     val missingFieldIds = requestedFieldIds.filterNot(declaredFieldIds.contains)
     if (missingFieldIds.nonEmpty) {
--- a/src/main/scala/read/MilvusPackedV2PartitionReader.scala
+++ b/src/main/scala/read/MilvusPackedV2PartitionReader.scala
@@ -71,6 +71,10 @@ object MilvusPackedV2PartitionReader {
       fieldMappings: FieldMappings,
       neededColumnFieldIds: Seq[Long]
   ): Array[String] = {
+    if (columnGroups.isEmpty) {
+      return Array.empty
+    }
+
     val declaredFieldIds = columnGroups.flatMap(_.fieldIds).toSet
     val requestedFieldIds: Seq[Long] =
       if (neededColumnFieldIds.nonEmpty) {
--- a/src/main/scala/read/MilvusPackedV2PartitionReader.scala
+++ b/src/main/scala/read/MilvusPackedV2PartitionReader.scala
@@ -24,6 +24,97 @@ import io.milvus.storage.{
   NativeLibraryLoader
 }
 
+object MilvusPackedV2PartitionReader {
+  private[read] case class FieldMappings(
+      fieldIdToName: Map[Long, String],
+      fieldNameToId: Map[String, Long],
+      fieldNameToArrowColumn: Map[String, String]
+  )
+
+  private[read] val SystemFieldAliases: Seq[(String, (Long, String))] = Seq(
+    "RowID" -> (0L, "RowID"),
+    "row_id" -> (0L, "RowID"),
+    "rowid" -> (0L, "RowID"),
+    "Timestamp" -> (1L, "Timestamp"),
+    "timestamp" -> (1L, "Timestamp")
+  )
+
+  private[read] def buildFieldMappings(
+      milvusSchema: CollectionSchema
+  ): FieldMappings = {
+    val systemFields = Map(0L -> "RowID", 1L -> "Timestamp")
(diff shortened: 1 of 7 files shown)


src/main/scala/read/MilvusPackedV2PartitionReader.scala
```diff
@@ -24,6 +24,97 @@ import io.milvus.storage.{
...
+        )
+        if (missingNames.nonEmpty) {
+          throw new IllegalArgumentException(
+            s"Packed V2 requested unknown columns: ${missingNames.distinct.mkString(",")}" +
+              s"; schema columns=${fieldMappings.fieldNameToId.keys.toSeq.sorted.mkString(",")}"
+          )
+        }
+        sourceSchema.fieldNames.toSeq.map(fieldMappings.fieldNameToId)
+      }
+    val missingFieldIds = requestedFieldIds.filterNot(declaredFieldIds.contains)
```
liliu-z, line 106: resolveNeededColumns throws for a V2 segment that has no column groups. V2SegmentLoader deliberately emits such segments (columnGroups = Seq.empty, V2SegmentLoader.scala:117) and they flow unfiltered into the reader, where an empty declaredFieldIds makes every requested field id 'missing', so the missingFieldIds.nonEmpty check (lines 99-108) throws IllegalArgumentException and aborts the whole Spark read task. The pre-PR reader filtered requested names against declared columns and returned an empty neededColumns array rather than throwing. Note the original 'dereferences the first file/row group' wording is inaccurate — there is no head/first deref; the abort is the missing-columns validation. Short-circuit the empty-columnGroups case and return an empty array.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -411,10 +411,10 @@ case class MilvusTable(
       }
     }
     if (fieldIDs.isEmpty || fieldIDs.contains("0")) {
-      fields = fields :+ StructField("row_id", LongType, false)
+      fields = fields :+ StructField("RowID", LongType, false)
```
liliu-z, line 422: MilvusTable.schema() appends RowID/Timestamp (lines 414, 417) with no check for a colliding user field, while MilvusSchemaUtil.convertToArrowSchema / convertToArrowSchemaWithFieldIdNames suppress the synthetic system field when a user field already uses an alias (case-insensitive name, or fieldID 0/1). For a collection whose fields are named e.g. row_id/timestamp, schema() advertises [RowID, Timestamp, row_id, timestamp, …] while the reader's Arrow schema collapses to [row_id, timestamp, …], so the schema shown to Spark disagrees with the columns the reader produces. Centralize the collision logic so schema() and SchemaUtil agree.

liliu-z, line 422: This renames the public output columns row_id→RowID and timestamp→Timestamp (lines 414, 417; also reflected in the snapshot test), a breaking change to the connector's output schema: existing jobs doing df.select("row_id") now fail to resolve, because the 'read-time aliases' only affect field-id projection, not how Spark resolves user-facing column names against the advertised schema. The commit is tagged enhance: rather than enhance!: (the repo's breaking-change marker, used by #97) and ships no migration notes. Mark it breaking and document the rename, ideally retaining row_id/timestamp as output aliases.

src/main/scala/read/MilvusPartitionReaderFactory.scala
```diff
@@ -51,7 +51,11 @@ class MilvusPartitionReaderFactory(
     with Logging {
 
   private def isSystemField(name: String): Boolean =
```
liliu-z: isSystemField is now dead code. This PR removed all three of its call sites (the v2Schema/innerSchema filters and both hasSyntheticFields checks) yet, in the same change, expanded the method body to also match RowID/Timestamp/rowid. The method is no longer referenced anywhere; remove it (the expansion is wasted).

src/main/scala/sources/MilvusDataSource.scala
congqixia: Thanks — that would be a breaking change only if released versions had
  already exposed working system fields under row_id/timestamp.

  The relevant baseline here is the released connector, not the partially
  wired internal names. In released behavior, system fields were not
  actually exposed as usable output columns: the reader factory stripped
  them from the inner schema and synthesized nulls, so there was no
  functional public contract for users to depend on.

  Given that baseline, this PR is not renaming an existing shipped feature;
  it is defining the first real public exposure of system fields, and it
  uses the underlying Milvus/snapshot canonical names RowID and Timestamp.

  So I agree this is a contract choice, but I don't think it is a breaking
  change relative to any released behavior. If we want lowercase aliases at
  the Spark schema surface, I see that as a separate compatibility/UX
  decision rather than preserving an already shipped contract.

src/main/scala/read/MilvusPackedV2PartitionReader.scala
```diff
@@ -24,6 +24,101 @@ import io.milvus.storage.{
...
+              s"; schema field IDs=${fieldMappings.fieldIdToName.keys.toSeq.sorted.mkString(",")}"
+          )
+        }
+        neededColumnFieldIds
+      } else {
+        val missingNames = sourceSchema.fieldNames.filterNot(
+          fieldMappings.fieldNameToId.contains
+        )
+        if (missingNames.nonEmpty) {
+          throw new IllegalArgumentException(
```
liliu-z, line 99: resolveNeededColumns throws `Packed V2 requested unknown columns: $meta` for any dynamic-field collection read through the packed-V2 path. neededColumnFieldIds is always empty here (set at MilvusDataSource.scala:1827), so the name branch runs and the synthetic `$meta` column (added to the schema at MilvusDataSource.scala:449) isn't in fieldNameToId, triggering the fail-fast. The previous reader and the V3 reader (MilvusLoonPartitionReader.getColumnNames) silently drop such columns so ArrowConverter nulls them. Recognize `$meta` as a tolerated dynamic column instead of aborting the read.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -410,11 +411,21 @@ case class MilvusTable(
...
-      fields = fields :+ StructField("row_id", LongType, false)
+    val missingSystemFields = MilvusSchemaUtil
+      .missingSystemFields(milvusCollection.schema)
+      .map(_.fieldID)
+      .toSet
+    if (
+      missingSystemFields.contains(0L) &&
+      (fieldIDs.isEmpty || fieldIDs.contains("0"))
+    ) {
+      fields = fields :+ StructField("RowID", LongType, false)
```
liliu-z, line 422: The RowID/Timestamp system-field definition is duplicated: SchemaUtil.scala:37-40 declares CanonicalSystemFields (RowID=0/Int64, Timestamp=1/Int64), while lines 422 and 428 re-hardcode the same fields as StructField("RowID", LongType, false)/StructField("Timestamp", LongType, false) rather than deriving them from that constant — even though this method already calls MilvusSchemaUtil.missingSystemFields. Each reader additionally redeclares the aliases (MilvusPackedV2PartitionReader.scala:34-40 and MilvusLoonPartitionReader SystemFieldAliases). These can drift in name/type/nullability; consolidate on one source of truth.

PR #99, opened 2026-06-04
enhance!: [reader] harden Arrow vector decoding

congqixia:
Use the existing `milvus.data_type` StructField metadata as the authoritative signal for FixedSizeBinary decoding so BinaryVector, Int8Vector, Float16Vector, and BFloat16Vector no longer depend on ambiguous Arrow physical shapes.

Align protobuf and snapshot schema mapping by exposing BinaryVector as Spark BinaryType and Int8Vector as Array[Short], then teach binary search and distance expressions to accept the new BinaryType contract without regressing legacy Array[Byte] callers.

### DIFF
--- a/src/main/scala/read/MilvusLoonPartitionReader.scala
+++ b/src/main/scala/read/MilvusLoonPartitionReader.scala
@@ -20,6 +20,7 @@ import org.apache.spark.sql.types.{
   LongType,
   ShortType,
   StringType,
+  StructField,
   StructType
 }
 
@@ -27,7 +28,7 @@ import com.zilliz.spark.connector.filter.VectorBruteForceSearch
 import com.zilliz.spark.connector.loon.Properties
 import com.zilliz.spark.connector.serde.ArrowConverter
 import com.zilliz.spark.connector.MilvusOption
-import io.milvus.grpc.schema.CollectionSchema
+import io.milvus.grpc.schema.{CollectionSchema, DataType}
 import io.milvus.storage.{
   ArrowUtils,
   LatestColumnGroupsResult,
@@ -64,6 +65,34 @@ object MilvusLoonPartitionReader {
       distance: Double,
       rowOffset: Long
   )
+
+  private[read] def decodeBinaryTypeVectorForSearch(
+      bytes: Array[Byte],
+      field: StructField
+  ): Array[Float] = {
+    if (!field.metadata.contains(ArrowConverter.MilvusDataTypeMetadataKey)) {
+      throw new IllegalArgumentException(
+        s"BinaryType vector search requires ${ArrowConverter.MilvusDataTypeMetadataKey} metadata"
+      )
+    }
+
+    field.metadata
+      .getLong(ArrowConverter.MilvusDataTypeMetadataKey)
+      .toInt match {
+      case DataType.FloatVector.value | DataType.Float16Vector.value |
+          DataType.BFloat16Vector.value =>
+        val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
+        (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
+      case DataType.BinaryVector.value =>
+        throw new IllegalArgumentException(
+          "BinaryVector does not support vector.search.* dense-float metrics; use binary search utilities with Hamming/Jaccard instead"
+        )
+      case other =>
+        throw new IllegalArgumentException(
+          s"BinaryType vector search is unsupported for Milvus data type $other"
+        )
+    }
+  }
 }
 
 // for Milvus 2.6+ version data source and milvus lake data
@@ -508,17 +537,16 @@ class MilvusLoonPartitionReader(
   ): Array[Float] = {
     dataType match {
       case ArrayType(FloatType, _) =>
-        // Array[Float] type
         val arrayData = row.getArray(colIndex)
         (0 until arrayData.numElements())
           .map(i => arrayData.getFloat(i))
           .toArray
 
       case BinaryType =>
-        // Binary type (for FixedSizeBinary float vectors)
-        val bytes = row.getBinary(colIndex)
-        val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
-        (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
+        MilvusLoonPartitionReader.decodeBinaryTypeVectorForSearch(
+          row.getBinary(colIndex),
+          sourceSchema(colIndex)
+        )
 
       case _ =>
         throw new IllegalArgumentException(
--- a/src/main/scala/read/MilvusLoonPartitionReader.scala
+++ b/src/main/scala/read/MilvusLoonPartitionReader.scala
@@ -24,10 +24,10 @@ import org.apache.spark.sql.types.{
   StructType
 }
 
+import com.zilliz.spark.connector.{FloatConverter, MilvusOption}
 import com.zilliz.spark.connector.filter.VectorBruteForceSearch
 import com.zilliz.spark.connector.loon.Properties
 import com.zilliz.spark.connector.serde.ArrowConverter
-import com.zilliz.spark.connector.MilvusOption
 import io.milvus.grpc.schema.{CollectionSchema, DataType}
 import io.milvus.storage.{
   ArrowUtils,
@@ -79,10 +79,19 @@ object MilvusLoonPartitionReader {
     field.metadata
       .getLong(ArrowConverter.MilvusDataTypeMetadataKey)
       .toInt match {
-      case DataType.FloatVector.value | DataType.Float16Vector.value |
-          DataType.BFloat16Vector.value =>
+      case DataType.FloatVector.value =>
         val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
         (0 until (bytes.length / 4)).map(_ => buffer.getFloat()).toArray
+      case DataType.Float16Vector.value =>
+        bytes
+          .grouped(2)
+          .map(b => FloatConverter.fromFloat16Bytes(b.toSeq))
+          .toArray
+      case DataType.BFloat16Vector.value =>
+        bytes
+          .grouped(2)
+          .map(b => FloatConverter.fromBFloat16Bytes(b.toSeq))
+          .toArray
       case DataType.BinaryVector.value =>
         throw new IllegalArgumentException(
           "BinaryVector does not support vector.search.* dense-float metrics; use binary search utilities with Hamming/Jaccard instead"
--- a/src/main/scala/read/MilvusLoonPartitionReader.scala
+++ b/src/main/scala/read/MilvusLoonPartitionReader.scala
@@ -66,6 +66,28 @@ object MilvusLoonPartitionReader {
       rowOffset: Long
   )
 
+  private[read] def validateVectorSearchField(
+      field: StructField,
+      metric: String
+  ): Unit = {
+    field.dataType match {
+      case BinaryType
+          if field.metadata.contains(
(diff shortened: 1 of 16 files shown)


src/main/scala/serde/ArrowConverter.scala
```diff
@@ -171,8 +200,23 @@ object ArrowConverter extends Logging {
         ArrayData.toArrayData(arrayElements)
 
       case BinaryType =>
-        val bytes = vector.asInstanceOf[VarBinaryVector].get(rowIndex)
-        bytes
+        vector match {
```
liliu-z, line 203: The `vector match` inside the `BinaryType` case enumerates only `VarBinaryVector` and `FixedSizeBinaryVector` with no `case _`, so any other Arrow vector mapped to `BinaryType` throws a raw `scala.MatchError`. This is reachable: `MilvusSnapshotReader` maps unrecognized Milvus types to `BinaryType` (`case _ => BinaryType`), and `FieldVector` is not sealed so the compiler emits no exhaustiveness warning. The `utf8StringFromVariableWidth` helper added in this same PR already includes a descriptive `case _ =>` fallback — add the same here so the error names the offending vector type instead of a bare MatchError.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -171,8 +200,27 @@ object ArrowConverter extends Logging {
...
+            v.get(rowIndex)
+          case v: FixedSizeBinaryVector =>
+            milvusType match {
+              case Some(MilvusDataType.BinaryVector) =>
+                v.get(rowIndex)
+              case Some(other) =>
+                throw new IllegalArgumentException(
+                  s"Cannot decode FixedSizeBinary as BinaryType for Milvus type $other"
+                )
+              case None =>
```
liliu-z, line 214: PR #99 makes FixedSizeBinary decoding of BinaryVector (now BinaryType) and Int8Vector (now ArrayType(ShortType)) require `milvus.data_type` metadata on the Spark field, throwing on `None`. MilvusBackfill's coalesce/overwrite source-read path builds its read schema via `MilvusSnapshotReader.fieldToSparkType` (MilvusBackfill.scala:294), which omits that metadata, and that schema flows unchanged through `.schema(withExtras)` → MilvusScan → the partition reader → `arrowToInternalRow`. So a backfill whose target field is a BinaryVector or Int8Vector now throws while reading the existing values to merge. Attach the `milvus.data_type` metadata when constructing those StructFields (as `DataTypeUtil.metadata`/`toSparkSchema` do). Scope is narrower than 'any backfill' and it fails on decode, not encode.

src/main/scala/serde/DataTypeUtil.scala
```diff
@@ -95,7 +95,7 @@ object DataTypeUtil {
       case MilvusDataType.FloatVector =>
         DataTypes.createArrayType(DataTypes.FloatType)
       case MilvusDataType.BinaryVector =>
-        DataTypes.createArrayType(DataTypes.BinaryType)
+        DataTypes.BinaryType
```
liliu-z, line 98: PR #99 remaps BinaryVector to Spark BinaryType here (and at MilvusSnapshotReader.scala:740), but the default insert writer was not updated: MilvusInsertDataWriter.scala:430-431 still reads a BinaryVector column with `row.getArray(fieldIndex).toByteArray()`, which is only valid for ArrayType. With a BinaryType column `getArray` misinterprets the raw binary payload (ClassCastException or corrupt array), so reading a Milvus collection and writing it back now breaks for BinaryVector. Read the column with `row.getBinary(fieldIndex)`. The adjacent Int8Vector branch stays correct because its reader type became ArrayType(ShortType), matching its `getArray().toShortArray()`.

src/main/scala/write/MilvusInsertDataWriter.scala
```diff
@@ -427,8 +427,7 @@ object MilvusInsertDataWriter {
         case MilvusDataType.BinaryVector =>
           buffer(fieldName)
             .asInstanceOf[ArrayBuffer[Array[Byte]]] += row
-            .getArray(fieldIndex)
-            .toByteArray()
+            .getBinary(fieldIndex)
```
liliu-z: The BinaryVector write path now reads the column with `row.getBinary(fieldIndex)` keyed only on the Milvus field type, with no check of the Spark column type. This is correct for the PR's new canonical `BinaryType` input, but a legacy `ArrayType(ByteType)` BinaryVector column on a production `UnsafeRow` makes `getBinary` return the serialized `UnsafeArrayData` blob (length + null bitmap + elements) verbatim, inserting a wrong-length, wrong-content vector with no exception; the new test only feeds `BinaryType` via `GenericInternalRow`, so it never exercises this. Unlike the search/distance paths this same PR updated to accept both forms (`byteArrayFromBinaryValue`, `extractByteArray`), the write path neither normalizes nor validates the column type. Dispatch on the Spark field type (getBinary for BinaryType, getArray().toByteArray() for ArrayType(ByteType)) or reject non-BinaryType explicitly; note that simply reverting to `getArray` as suggested would instead corrupt the canonical BinaryType input.

src/main/scala/serde/ArrowConverter.scala
liliu-z, line 214: In snapshot mode `MilvusTable.schema()` returns a user-supplied external schema verbatim (`MilvusDataSource.scala:397-401`, enabled by `supportsExternalMetadata = true` at :154 — appendExtraColumns only appends, it does not re-inject field metadata). Vector fields the caller did not tag with `milvus.data_type` therefore reach the executor without metadata, so `arrowToInternalRow` derives `milvusType = None` and the FixedSizeBinary decode throws on the first non-null row for BinaryVector (BinaryType, here at :214), Int8Vector (Array[Short], :153), and the legacy Array[Byte] path (:137). This regresses snapshot reads that previously passed their own `.schema()`: pre-PR the Array[Byte]/FixedSizeBinary path decoded without any metadata requirement. The reader already receives the authoritative `milvusSchema`; re-inject the metadata from the Milvus FieldSchema (e.g. via `fieldToStructField`/`DataTypeUtil.metadata`) when an external schema is supplied.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -329,6 +331,35 @@ case class MilvusTable(
 
   override def name(): String = milvusOption.collectionName
 
+  private def rehydrateSnapshotSchemaMetadata(
+      baseSchema: StructType
+  ): StructType = {
+    val fieldTypeByName = milvusCollection.schema.fields.map { field =>
```
liliu-z, line 352: rehydrateSnapshotSchemaMetadata sources the injected milvus.data_type values from milvusCollection.schema, but initFromSnapshot populates that schema only from SnapshotSchemaJson (line 224) and leaves it empty when the snapshot carries its schema via SnapshotSchemaBytes — the format MilvusBackfill (line 608) and buildClientSnapshotOptions (line 1225) actually emit. On MilvusBackfill's spark.read.schema(...).load() path this makes the rehydration a silent no-op; it is currently harmless only because that schema is built with fieldToStructField, which already embeds the metadata. Any external-schema snapshot read that relies on SnapshotSchemaBytes without per-field metadata would leave FixedSizeBinary vector columns unhydrated and fail to decode. Have rehydration (or initFromSnapshot) also parse SnapshotSchemaBytes.

src/main/scala/serde/DataTypeUtil.scala
```diff
@@ -95,7 +95,7 @@ object DataTypeUtil {
       case MilvusDataType.FloatVector =>
         DataTypes.createArrayType(DataTypes.FloatType)
       case MilvusDataType.BinaryVector =>
```
liliu-z, line 97: Mapping BinaryVector to Spark BinaryType (here and at MilvusSnapshotReader.scala:744) routes binary vectors into the pre-existing `case BinaryType` branch of MilvusLoonPartitionReader.extractVectorFromRow (line 517), which reinterprets the raw bytes as little-endian floats. Previously BinaryVector mapped to an array type that fell through to the `case _` throw, so the row was caught-and-skipped; now the built-in vector search (vector.search.* with an L2/IP/COSINE metric) pointed at a binary column silently yields garbage distances and wrong results with no error. extractVectorFromRow is unchanged by this PR, so special-case BinaryType there to reject (or correctly handle via Hamming/Jaccard) binary vectors.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -218,14 +220,31 @@ case class MilvusTable(
...
+      .flatMap { json =>
+        MilvusSnapshotReader.parseSnapshotMetadata(json) match {
+          case Right(metadata) =>
+            Try {
+              CollectionSchema.parseFrom(
+                MilvusSnapshotReader.toProtobufSchemaBytes(
+                  metadata.collection.schema
+                )
+              )
+            }.toOption
```
liliu-z: The protobuf decode of SnapshotSchemaJson (228-234) and the new SnapshotSchemaBytes path (241-245) are both wrapped in `Try{...}.toOption`, so a malformed value becomes None and createMinimalCollectionSchema falls back to an empty CollectionSchema (269-274), indistinguishable from no snapshot schema being supplied. Before this PR the JSON path's `parseFrom` was unwrapped and threw on bad input; this change newly swallows it. With an empty schema, rehydrateSnapshotSchemaMetadata finds no fields and injects no `milvus.data_type` metadata, so the read later fails inside ArrowConverter with a generic 'requires metadata' error that never names the offending option. Let the parse exception propagate (including the option name) instead of degrading to an empty schema.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -64,6 +65,34 @@ object MilvusLoonPartitionReader {
...
+    if (!field.metadata.contains(ArrowConverter.MilvusDataTypeMetadataKey)) {
+      throw new IllegalArgumentException(
+        s"BinaryType vector search requires ${ArrowConverter.MilvusDataTypeMetadataKey} metadata"
+      )
+    }
+
+    field.metadata
+      .getLong(ArrowConverter.MilvusDataTypeMetadataKey)
+      .toInt match {
+      case DataType.FloatVector.value | DataType.Float16Vector.value |
```
liliu-z: decodeBinaryTypeVectorForSearch groups Float16Vector and BFloat16Vector into the same branch as FloatVector and decodes them with 4-byte float logic (`bytes.length / 4`, `buffer.getFloat()`), but both half-precision formats are 2 bytes per element — this would return half the expected element count with garbage values. The canonical decoder ArrowConverter.decodeFixedSizeBinaryFloats correctly decodes them via `bytes.grouped(2)` and FloatConverter.fromFloat16Bytes/fromBFloat16Bytes. The branch is currently unreachable because arrowToInternalRow runs first and rejects a BinaryType FixedSizeBinary column carrying Float16/BFloat16 metadata, but the decode is wrong and misleading: either decode at 2-byte width or reject these two types here the way BinaryVector is already rejected.

src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -64,6 +65,43 @@ object MilvusLoonPartitionReader {
...
+        bytes
+          .grouped(2)
+          .map(b => FloatConverter.fromFloat16Bytes(b.toSeq))
+          .toArray
+      case DataType.BFloat16Vector.value =>
+        bytes
+          .grouped(2)
+          .map(b => FloatConverter.fromBFloat16Bytes(b.toSeq))
+          .toArray
+      case DataType.BinaryVector.value =>
```
liliu-z, line 117: The IllegalArgumentException raised for a BinaryVector dense search (line 96) is reached through extractVectorFromRow and caught by the `catch { case e: Exception => logWarning(...); null }` in processBatch (lines 454-466). The rejection never propagates: a dense vector.search against a BinaryVector column drops every row as a null vector and returns empty results while logging a warning per row. Validate the unsupported metric up front, or let this rejection escape processBatch instead of catching it.

PR #100, opened 2026-06-09
enhance!: [storage] harden IAM storage option validation

congqixia:
## Summary
- require `fs.bucket_name` when `fs.use_iam=true` in the shared storage-property builder
- stop forwarding empty static AK/SK in IAM mode so the native layer can use the AWS default credentials chain
- normalize forwarded storage options such as `fs.use_iam` and `fs.bucket_name` before handing them to the FFI
- keep the configurable bytes-per-value override for variable-width Arrow columns

## Breaking change
- this change affects both write and read paths because they share `Properties.fromMilvusOption`
- when `fs.use_iam=true`, the connector now fails fast if `fs.bucket_name` is missing instead of silently falling back to `a-bucket`
- workloads that relied on that implicit fallback must set `fs.bucket_name` explicitly

## Test plan
- [x] `sbt "testOnly com.zilliz.spark.connector.loon.PropertiesTest -- -oD"`
- [x] existing writer/read tests for the touched code paths

### DIFF
--- a/docs/reference-en.md
+++ b/docs/reference-en.md
@@ -139,6 +139,7 @@ val s3Options = Map(
 | Parameter | Type | Required | Default | Description |
 |-----------|------|----------|---------|-------------|
 | `MilvusOption.MilvusInsertMaxBatchSize` | Int | No | 5000 | Maximum batch size for single insert operation |
+| `MilvusOption.WriterVariableWidthBytesPerValue` | Double | No | 32.0 | Initial Arrow buffer density for VARCHAR/JSON/binary writer columns; increase for wide variable-width values. Must be finite and positive. |
 | `MilvusOption.MilvusRetryCount` | Int | No | 3 | Number of retries on operation failure |
 | `MilvusOption.MilvusRetryInterval` | Int | No | 1000 | Retry interval in milliseconds |
 
--- a/src/main/scala/loon/Properties.scala
+++ b/src/main/scala/loon/Properties.scala
@@ -13,6 +13,11 @@ object Properties {
       options: scala.collection.Map[String, String]
   ): Option[String] = options.get(FsConfig.FsUseIam).map(_.trim)
 
+  private[connector] def normalizedFsBucketNameValue(
+      options: scala.collection.Map[String, String]
+  ): Option[String] =
+    options.get(FsConfig.FsBucketName).map(_.trim).filter(_.nonEmpty)
+
   /** Filesystem configuration constants for Storage V2 (matching milvus-storage
     * C++ API)
     */
@@ -58,9 +63,7 @@ object Properties {
     // dev/test default when we are clearly *not* in IAM mode.
     val normalizedUseIamValue = normalizedFsUseIamValue(milvusOption.options)
     val useIamFlag = normalizedUseIamValue.exists(_.equalsIgnoreCase("true"))
-    val bucket = milvusOption.options
-      .get(FsConfig.FsBucketName)
-      .filter(_.trim.nonEmpty)
+    val bucket = normalizedFsBucketNameValue(milvusOption.options)
       .getOrElse {
         if (useIamFlag) {
           throw new IllegalArgumentException(
--- a/src/main/scala/loon/Properties.scala
+++ b/src/main/scala/loon/Properties.scala
@@ -63,14 +63,25 @@ object Properties {
     // dev/test default when we are clearly *not* in IAM mode.
     val normalizedUseIamValue = normalizedFsUseIamValue(milvusOption.options)
     val useIamFlag = normalizedUseIamValue.exists(_.equalsIgnoreCase("true"))
+    val rawBucketValue = milvusOption.options.get(FsConfig.FsBucketName)
     val bucket = normalizedFsBucketNameValue(milvusOption.options)
       .getOrElse {
-        if (useIamFlag) {
-          throw new IllegalArgumentException(
-            s"${FsConfig.FsBucketName} must be set when ${FsConfig.FsUseIam}=true"
-          )
+        rawBucketValue match {
+          case Some(_) if useIamFlag =>
+            throw new IllegalArgumentException(
+              s"${FsConfig.FsBucketName} must not be blank when ${FsConfig.FsUseIam}=true"
+            )
+          case Some(_) =>
+            throw new IllegalArgumentException(
+              s"${FsConfig.FsBucketName} must not be blank"
+            )
+          case None if useIamFlag =>
+            throw new IllegalArgumentException(
+              s"${FsConfig.FsBucketName} must be set when ${FsConfig.FsUseIam}=true"
+            )
+          case None =>
+            "a-bucket"
         }
-        "a-bucket"
       }
     val rootPath = milvusOption.options.getOrElse(FsConfig.FsRootPath, "files")
     val accessKey =
--- a/src/main/scala/loon/Properties.scala
+++ b/src/main/scala/loon/Properties.scala
@@ -42,15 +42,11 @@ object Properties {
     *   MilvusStorageProperties for milvus-storage native library
     */
   def fromMilvusOption(milvusOption: MilvusOption): MilvusStorageProperties = {
-    val props = new MilvusStorageProperties()
     val propsMap = new ju.HashMap[String, String]()
 
     // Extract filesystem configuration with defaults
     val endpoint =
       milvusOption.options.getOrElse(FsConfig.FsAddress, "localhost:9000")
-    val bucket =
-      milvusOption.options.getOrElse(FsConfig.FsBucketName, "a-bucket")
-    val rootPath = milvusOption.options.getOrElse(FsConfig.FsRootPath, "files")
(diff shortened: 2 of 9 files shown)


src/main/scala/loon/Properties.scala
```diff
@@ -59,6 +55,18 @@ object Properties {
     val useIamFlag = milvusOption.options
       .get(FsConfig.FsUseIam)
       .exists(_.equalsIgnoreCase("true"))
```
liliu-z: `useIamFlag` is parsed with `.exists(_.equalsIgnoreCase("true"))` and no `.trim`, so `fs.use_iam = "true "` (trailing whitespace, e.g. from a properties file or env interpolation) evaluates to false. That silently disables IAM mode: the new `fs.bucket_name` fail-fast is skipped and the `minioadmin` AK/SK defaults are injected, so the write falls back to the `a-bucket` default — exactly the footgun this PR sets out to close. Trim before the comparison. Note the bucket lookup just below already trims via `.filter(_.trim.nonEmpty)`, so the whitespace handling is internally inconsistent.

docs/reference-en.md
```diff
@@ -139,6 +139,7 @@ val s3Options = Map(
 | Parameter | Type | Required | Default | Description |
 |-----------|------|----------|---------|-------------|
 | `MilvusOption.MilvusInsertMaxBatchSize` | Int | No | 5000 | Maximum batch size for single insert operation |
+| `MilvusOption.WriterVariableWidthBytesPerValue` | Double | No | 32.0 | Initial Arrow buffer density for VARCHAR/JSON/binary writer columns; increase for wide variable-width values. Must be finite and positive. |
```
liliu-z, line 142: `WriterVariableWidthBytesPerValue` is documented here as a general write parameter, but only `MilvusLoonPartitionWriter` reads it; `MilvusV2BinlogWriter` still hardcodes `32.0` and never consults the option, so it silently no-ops for the binlog write path. Either qualify the doc as loon-writer-only or wire the option into the binlog writer.

src/main/scala/loon/Properties.scala
```diff
@@ -42,23 +42,31 @@ object Properties {
...
-    val rootPath = milvusOption.options.getOrElse(FsConfig.FsRootPath, "files")
     // When use_iam=true the FFI must consult the AWS default credentials
     // chain (env vars / web identity / instance profile). Falling back to
     // hard-coded "minioadmin" defaults under IRSA produces signed requests
     // with bogus keys, which fail with InvalidAccessKeyId. Only inject the
     // dev/test default when we are clearly *not* in IAM mode.
     val useIamFlag = milvusOption.options
       .get(FsConfig.FsUseIam)
-      .exists(_.equalsIgnoreCase("true"))
+      .exists(_.trim.equalsIgnoreCase("true"))
```
liliu-z: The PR trims use_iam for the IAM decision at :57 (.exists(_.trim.equalsIgnoreCase("true"))) but the value forwarded to the native FFI at :111-113 is still the raw, untrimmed string. So use_iam=" true " enables IAM on the JVM side (blanking AK/SK and forcing a bucket) while the native layer receives " true "; if the native parser matches exactly, the two layers disagree about whether IAM is active. Forward the trimmed value so both sides stay consistent.

src/main/scala/write/MilvusV2BinlogWriter.scala
```diff
@@ -174,6 +185,8 @@ class MilvusV2BinlogWriter(
   private val batchSize: Int =
     if (milvusOption.insertMaxBatchSize > 0) milvusOption.insertMaxBatchSize
     else 5000
+  private val variableWidthBytesPerValue: Double =
```
liliu-z: Option validation at line 188 runs only after `nativeProperties`, `arrowSchemaC`, and the native writer have already been opened (lines 144–170). Because validation throws from the constructor and the constructor never returns, the object is never assigned and `abort()` can never run, so a bad option value leaks the native allocations on every failed construction — a hazard the author's own comment at :162-163 names explicitly. Move the validation before the native allocations, as the loon writer already does.

src/main/scala/loon/Properties.scala
```diff
@@ -42,23 +46,30 @@ object Properties {
...
-      .get(FsConfig.FsUseIam)
-      .exists(_.equalsIgnoreCase("true"))
+    val normalizedUseIamValue = normalizedFsUseIamValue(milvusOption.options)
+    val useIamFlag = normalizedUseIamValue.exists(_.equalsIgnoreCase("true"))
+    val bucket = milvusOption.options
+      .get(FsConfig.FsBucketName)
+      .filter(_.trim.nonEmpty)
+      .getOrElse {
+        if (useIamFlag) {
+          throw new IllegalArgumentException(
```
liliu-z: The new IAM guard lives in `fromMilvusOption`, which both read paths invoke (`MilvusLoonPartitionReader.scala:196`, `MilvusPackedV2PartitionReader.scala:202`), so a read with `fs.use_iam=true` and no `fs.bucket_name` now throws where it previously fell back to the `a-bucket` default. That is a read-facing behavior change, but the PR is titled `enhance: [writer]` (non-breaking) — the comparable reader guard in #99 shipped as `enhance!:`. Classify this as breaking and document the read-path impact; no code change needed.

src/main/scala/loon/Properties.scala
```diff
@@ -42,23 +46,30 @@ object Properties {
...
     // with bogus keys, which fail with InvalidAccessKeyId. Only inject the
     // dev/test default when we are clearly *not* in IAM mode.
-    val useIamFlag = milvusOption.options
-      .get(FsConfig.FsUseIam)
-      .exists(_.equalsIgnoreCase("true"))
+    val normalizedUseIamValue = normalizedFsUseIamValue(milvusOption.options)
+    val useIamFlag = normalizedUseIamValue.exists(_.equalsIgnoreCase("true"))
+    val bucket = milvusOption.options
+      .get(FsConfig.FsBucketName)
+      .filter(_.trim.nonEmpty)
```
liliu-z: `bucket` is filtered on `_.trim.nonEmpty` but the original un-trimmed value is what gets forwarded at `propsMap.put(FsConfig.FsBucketName, bucket)` (L95), so a value like `"my-bucket "` passes validation and reaches the native layer with trailing whitespace, failing deep in the FFI with an opaque error. Its sibling `fs.use_iam` is normalized with `.map(_.trim)` (via `normalizedFsUseIamValue`, L114); apply the same to the bucket — `.map(_.trim).filter(_.nonEmpty)`.

src/main/scala/loon/Properties.scala
congqixia: Good catch. This IAM guard is shared by the read path as well, so it is a breaking behavior change rather than a writer-only hardening. I'll update the PR classification/description to call out the read-path impact explicitly.

src/main/scala/loon/Properties.scala
```diff
@@ -42,23 +51,28 @@ object Properties {
...
+    val normalizedUseIamValue = normalizedFsUseIamValue(milvusOption.options)
+    val useIamFlag = normalizedUseIamValue.exists(_.equalsIgnoreCase("true"))
+    val bucket = normalizedFsBucketNameValue(milvusOption.options)
+      .getOrElse {
+        if (useIamFlag) {
+          throw new IllegalArgumentException(
+            s"${FsConfig.FsBucketName} must be set when ${FsConfig.FsUseIam}=true"
+          )
+        }
+        "a-bucket"
```
liliu-z: In non-IAM mode a missing, blank, or whitespace-only `fs.bucket_name` is silently substituted with the placeholder `a-bucket` (line 73) rather than rejected. Because `normalizedFsBucketNameValue` now filters blank values to None, this PR routes present-but-empty bucket values to the placeholder too (previously they were forwarded as-is and failed at props creation), so a mistyped/empty bucket proceeds against `a-bucket` and surfaces only later as a confusing not-found/access-denied error. Consider failing fast (as the IAM path now does) or at least logging a warning for a blank non-IAM bucket.

src/main/scala/write/MilvusLoonWriter.scala
```diff
@@ -218,6 +242,12 @@ class MilvusLoonPartitionWriter(
 
   // Batch size configuration
   private val batchSize = milvusOption.insertMaxBatchSize
+  private val variableWidthBytesPerValue =
```
liliu-z: The new `variableWidthBytesPerValue` field is validated in the constructor at line 245 — after `allocator` (215) and `root` (241) are already created — and `parsePositiveDoubleOption` throws on an invalid value. Since `cleanup()` only runs via commit/abort/close on a fully-constructed writer, an invalid config value throws out of the constructor and leaks the RootAllocator/VectorSchemaRoot (the same uncaught-throw pattern also leaks them when the new IAM-bucket check inside the `fromMilvusOption` call at line 274 fires). `MilvusV2BinlogWriter` avoids this by validating the same option before allocating its allocator (line 92 vs 122); reorder the V3 writer to validate first, or wrap construction in try/catch.

PR #101, opened 2026-06-11
enhance: Reject flags as backfill option values

congqixia:
Fail fast when a key/value backfill option is immediately followed by another flag. This reports a missing value at parse time instead of mis-parsing the next flag and failing later with a confusing positional-argument error.

### DIFF
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -218,7 +218,11 @@ object BackfillApp {
             i += 1
           } else if (i + 1 < args.length) {
             val value = args(i + 1)
-            if (value.startsWith("--")) {
+            if (
+              value.startsWith("--") && KnownFlags.contains(
+                value.stripPrefix("--")
+              )
+            ) {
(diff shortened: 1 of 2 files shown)


src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -217,7 +217,11 @@ object BackfillApp {
             map += (key -> "true")
             i += 1
           } else if (i + 1 < args.length) {
-            map += (key -> args(i + 1))
+            val value = args(i + 1)
+            if (value.startsWith("--")) {
```
liliu-z: The new `if (value.startsWith("--"))` guard rejects any backfill option value beginning with `--`, not just misplaced flags, so a storage secret or other value that legitimately starts with `--` can no longer be passed. In practice none of the 16 KV flags (paths, endpoints, AWS keys, region, mode) takes such a value, and this closes a real silent mis-parse where e.g. `--source-s3-endpoint --source-use-iam` previously stored `--source-use-iam` as the endpoint; single-dash values still work, so this is a deliberate low-risk narrowing worth confirming no downstream invocation relies on a `--`-prefixed value.

PR #102, opened 2026-07-03
enhance: [reader] add StorageV2 deltalog planning and delete-aware reads

congqixia:
Add StorageV2 deltalog propagation, snapshot/client-snapshot planning, and PK-delete filtering on the packed-V2 read path, while keeping backfill explicitly delete-blind behind separate read/backfill options.

This change carries StorageV2 delta metadata through manifest/snapshot models, builds per-segment delete plans in Scala before reader creation, and applies them in packed-V2 readers without renumbering original row offsets. It also hardens client-snapshot bucket resolution for relative paths, blocks the legacy non-snapshot fallback on StorageV2 segments, and cleans up stale logging around ignored delete metadata.

Current scope is intentionally limited: only StorageV2 deltalogs are handled here, only PK-delete semantics are applied, and StorageV3 manifest-based deltalogs remain unsupported. StorageV2 delete-only/L0 segments are skipped when delete application is disabled rather than being materialized into backfill-style reads.

### DIFF
--- a/src/main/scala/read/MilvusDeltaLogReader.scala
+++ b/src/main/scala/read/MilvusDeltaLogReader.scala
@@ -60,15 +60,17 @@ object MilvusDeltaLogReader extends Logging {
     )
   }
 
+  private val AllPartitionsId = -1L
+
   private[read] def mergeInheritedDeletePlans(
       dataSegments: Seq[V2SegmentInfo],
       inheritedPlansByPartition: Map[Long, MilvusDeletePlan],
       ownPlansBySegment: Map[Long, MilvusDeletePlan]
   ): Map[Long, MilvusDeletePlan] = {
     dataSegments.iterator.map { seg =>
-      val inheritedPlan = inheritedPlansByPartition.getOrElse(
+      val inheritedPlan = effectiveInheritedDeletePlan(
         seg.partitionId,
-        MilvusDeletePlan.empty
+        inheritedPlansByPartition
       )
       val ownPlan = ownPlansBySegment.getOrElse(
         seg.segmentId,
@@ -96,6 +98,31 @@ object MilvusDeltaLogReader extends Logging {
       }
     ).map(_.toMap)
   }
+
+  private[read] def effectiveInheritedDeletePlan(
+      partitionId: Long,
+      inheritedPlansByPartition: Map[Long, MilvusDeletePlan]
+  ): MilvusDeletePlan = {
+    val collectionWidePlan = inheritedPlansByPartition.getOrElse(
+      AllPartitionsId,
+      MilvusDeletePlan.empty
+    )
+    val partitionPlan = inheritedPlansByPartition.getOrElse(
+      partitionId,
+      MilvusDeletePlan.empty
+    )
+    MilvusDeletePlan.union(collectionWidePlan, partitionPlan)
+  }
+
+  private[read] def inheritedDeletePlanPartitionMarker(
+      partitionId: Long,
+      inheritedPlansByPartition: Map[Long, MilvusDeletePlan]
+  ): Option[Long] =
+    if (
+      inheritedPlansByPartition.contains(AllPartitionsId) ||
+      inheritedPlansByPartition.contains(partitionId)
+    ) Some(partitionId)
+    else None
 
   def loadDeletePlan(
       deltaLogs: Seq[V2DeltaLogFile],
--- a/src/main/scala/read/MilvusDeltaLogReader.scala
+++ b/src/main/scala/read/MilvusDeltaLogReader.scala
@@ -40,21 +40,61 @@ object MilvusDeltaLogReader extends Logging {
     val dataSegments = segments.filter(_.columnGroups.nonEmpty)
 
     for {
-      globalPlan <- loadDeletePlan(
-        deleteOnlySegments.flatMap(_.deltaLogs),
+      globalPlans <- loadPartitionScopedDeletePlans(
+        deleteOnlySegments,
         pkField,
         bucket,
         hadoopConf
       )
-      perSegment <- sequence(
+      ownPlans <- sequence(
         dataSegments.map { seg =>
           loadDeletePlan(seg.deltaLogs, pkField, bucket, hadoopConf).map {
-            ownPlan =>
-              seg.segmentId -> MilvusDeletePlan.union(globalPlan, ownPlan)
+            ownPlan => seg.segmentId -> ownPlan
           }
         }
       )
-    } yield perSegment.toMap
+    } yield mergeInheritedDeletePlans(
+      dataSegments,
+      globalPlans,
+      ownPlans.toMap
+    )
+  }
+
+  private[read] def mergeInheritedDeletePlans(
+      dataSegments: Seq[V2SegmentInfo],
+      inheritedPlansByPartition: Map[Long, MilvusDeletePlan],
+      ownPlansBySegment: Map[Long, MilvusDeletePlan]
+  ): Map[Long, MilvusDeletePlan] = {
+    dataSegments.iterator.map { seg =>
+      val inheritedPlan = inheritedPlansByPartition.getOrElse(
+        seg.partitionId,
+        MilvusDeletePlan.empty
+      )
+      val ownPlan = ownPlansBySegment.getOrElse(
+        seg.segmentId,
+        MilvusDeletePlan.empty
+      )
+      seg.segmentId -> MilvusDeletePlan.union(inheritedPlan, ownPlan)
+    }.toMap
+  }
+
+  private[connector] def loadPartitionScopedDeletePlans(
+      deleteOnlySegments: Seq[V2SegmentInfo],
+      pkField: FieldSchema,
+      bucket: String,
+      hadoopConf: Configuration
+  ): Either[Throwable, Map[Long, MilvusDeletePlan]] = {
+    sequence(
+      deleteOnlySegments.groupBy(_.partitionId).toSeq.map {
+        case (partitionId, segments) =>
+          loadDeletePlan(
+            segments.flatMap(_.deltaLogs),
+            pkField,
+            bucket,
+            hadoopConf
+          ).map(partitionId -> _)
+      }
+    ).map(_.toMap)
   }
 
   def loadDeletePlan(
@@ -99,8 +139,8 @@ object MilvusDeltaLogReader extends Logging {
       pkField: FieldSchema,
       path: String
   ): MilvusDeletePlan = {
-    val longs = mutable.HashSet.empty[Long]
-    val strings = mutable.HashSet.empty[String]
+    val longs = mutable.HashMap.empty[Long, Long]
+    val strings = mutable.HashMap.empty[String, Long]
     val reader = newGroupReader(payload)
     try {
       var record = reader.read()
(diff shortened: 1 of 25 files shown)


src/main/scala/tools/ReadSourceOnlyApp.scala
```diff
@@ -0,0 +1,306 @@
...
+        options = options + (MilvusOption.SnapshotV2Segments ->
+          MilvusSnapshotReader.serializeV2Segments(v2Segments))
+      }
+
+      // 5. Build the read schema: PK field + segment metadata columns
+      import org.apache.spark.sql.types._
+      val segmentIdCol = MilvusOption.MilvusExtraColumnSegmentID
+      val rowOffsetCol = MilvusOption.MilvusExtraColumnRowOffset
+      val pkStructField = MilvusSnapshotReader.fieldToStructField(pkField)
+      val readSchema = StructType(
```
liliu-z, line 150: The app adds `$segment_id`/`$row_offset` to `readSchema` (L157-169) but its `options` map (L118-150) never sets `milvus.extra.columns`, and `isMetadataExtraField` (`MilvusPartitionReaderFactory.scala:39`) strips metadata fields only via `requestedExtraColumns.contains(name)` — where `requestedExtraColumns` (L18-32) is built solely from that option — so with it unset these two fields are never removed from `innerSchema` and flow into `resolveNeededColumns`, which recognizes them in neither `fieldNameToId` nor `ToleratedUnmappedColumns` (only `$meta`). Consequently `df.count()` throws `Packed V2 requested unknown columns: $segment_id,$row_offset` at runtime, so the diagnostic app fails loudly instead of reading. Set `milvus.extra.columns` to include these fields so they route through the metadata-injection wrapper rather than being requested as real columns.

src/main/scala/read/MilvusSegmentManifestReader.scala
```diff
@@ -60,33 +62,28 @@ private[read] case class AvroManifestEntry(
...
-    * writer schema exactly for whatever it parses. By stopping at
-    * `binlog_files` (the last field backfill actually uses) we make all
-    * trailing fields — present or absent — effectively optional: extra bytes
-    * after this field are simply ignored, and writers that emit fewer trailing
-    * fields still decode because the prefix they share with us is identical.
-    * Update this if backfill starts consuming a later field.
-    */
-  private val LastNeededField = "binlog_files"
+  private val SchemaResources: Map[Int, String] = Map(
+    1 -> "/milvus-segment-manifest-v1.avsc",
```
liliu-z, line 68: The reader now loads AVRO schemas from `/milvus-segment-manifest-v1.avsc`..`-v4.avsc`, but none of these four resources exist — only the old `src/main/resources/milvus-segment-manifest.avsc` is present and no build step generates the versioned copies. Because `schemas` is a lazy val that eagerly maps over all four resources, the first StorageV2 manifest parse — which happens on every backfill and every snapshot V2 read via `V2SegmentLoader.loadV2Segments` → `MilvusSegmentManifestReader.parse` (even at the default version 1) — forces `loadSchema` and throws `bundled AVSC resource /milvus-segment-manifest-v1.avsc not found on classpath`, breaking all StorageV2 reads. Add the versioned AVSC files (the old file already contains `segment_level` and `deltalog_files`).

src/main/scala/read/MilvusDeltaLogReader.scala
```diff
@@ -0,0 +1,475 @@
...
+      )
+    }
+    if (record.getFieldRepetitionCount(0) == 0) {
+      throw new IllegalStateException(
+        s"multi-field delete log payload in $path is missing the pk value"
+      )
+    }
+
+    pkField.dataType match {
+      case DataType.Int64   => longs += record.getLong(0, 0)
```
liliu-z: Delete application is timestamp-blind. appendMultiFieldDelete guards `fieldCount < 2` to confirm a ts column exists but reads only the pk at column 0; MilvusDeletePlan stores just Set[Long]/Set[String] with no timestamp; and MilvusPackedV2PartitionReader.isDeleted drops any row whose pk is in the plan regardless of the row's insert ts. This diverges from Milvus MVCC, where a delete at ts T only masks rows with ts <= T. On the common upsert/reinsert pattern (delete pk, then insert the same pk at a newer ts), the delete lands in an L0 segment whose plan is unioned into the reinserted row's segment, so the fresh row is wrongly dropped. Delete application is on by default (milvus.read.apply.deletes defaults true), so reads silently return stale results. Carry the delete ts through and filter by (pk, ts).

src/main/scala/read/MilvusDeltaLogReader.scala
```diff
@@ -0,0 +1,475 @@
...
+        deleteOnlySegments.flatMap(_.deltaLogs),
+        pkField,
+        bucket,
+        hadoopConf
+      )
+      perSegment <- sequence(
+        dataSegments.map { seg =>
+          loadDeletePlan(seg.deltaLogs, pkField, bucket, hadoopConf).map {
+            ownPlan =>
+              seg.segmentId -> MilvusDeletePlan.union(globalPlan, ownPlan)
```
liliu-z: globalPlan is built from every delete-only/L0 segment's deltalogs across all partitions (deleteOnlySegments.flatMap(_.deltaLogs)) and then unioned into each data segment's plan without consulting partitionId, even though V2SegmentInfo.partitionId is available. A partition-scoped L0 delete of pk X therefore also masks a distinct row with pk X living in another partition. Impact is narrower than the timestamp gap: collection-wide pk deletes are handled correctly by the global union, so this only over-deletes when the same pk exists as separate rows in multiple partitions and the delete was partition-scoped. Still, the partition scope is silently discarded; scope L0/global deletes to their owning partition.

liliu-z: union(globalPlan, ownPlan) copies the full L0 delete set into every data segment's plan, and buildSnapshotPartitions stores each segment's plan as a field on its MilvusPackedV2InputPartition, which is serialized into the corresponding Spark task. The L0 delete set is thus duplicated once per data segment, so total serialized task size grows with (#data-segments x #L0-deletes) instead of being shared. On collections with large L0 delete sets and many segments this bloats task binaries. Consider broadcasting a single shared delete set rather than embedding a per-segment copy.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1798,100 +2001,140 @@ class MilvusScan(
...
-        .map { json =>
-          MilvusSnapshotReader.deserializeV2Segments(json) match {
-            case Right(segs) if segs.nonEmpty =>
-              logInfo(
-                s"Creating ${segs.size} packed-V2 partition(s) from " +
-                  s"SnapshotV2Segments option"
+  override def createReaderFactory(): PartitionReaderFactory = {
+    val optionsMap = options.asScala.toMap
+    val inheritedPlansByPartition =
+      if (
```
liliu-z, line 2100: On the client-snapshot fast path the `isSnapshotMode(options)` gate at line 2079 is statically false (per `MilvusOption.isSnapshotModeFrom`), so `createReaderFactory` builds an empty inherited-delete context while partitions carry only the `inheritedDeletePlanPartitionId` marker (`:1924`). Own-segment deletes still apply, but inherited L0/uncompacted deletes are silently dropped, so a default read returns rows that collection-level or L0 deletes had already removed. Populate the inherited-delete context on the fast path, e.g. key the gate off `readApplyDeletes` rather than `isSnapshotMode`.

src/main/scala/read/MilvusDeltaLogReader.scala
```diff
@@ -0,0 +1,545 @@
...
+  }
+
+  private[connector] def loadPartitionScopedDeletePlans(
+      deleteOnlySegments: Seq[V2SegmentInfo],
+      pkField: FieldSchema,
+      bucket: String,
+      hadoopConf: Configuration
+  ): Either[Throwable, Map[Long, MilvusDeletePlan]] = {
+    sequence(
+      deleteOnlySegments.groupBy(_.partitionId).toSeq.map {
```
liliu-z, line 90: Deltalogs are matched to data segments by exact partition-ID equality, but Milvus writes collection-wide deletes (delete-by-expr with no partition filter) under the sentinel `partition == AllPartitionsID` (`-1`), and upstream's querynode L0 path loads the `-1` deltalogs for every target partition. The exact-match grouping (also `MilvusPartitionReaderFactory.scala:146`) never joins key `-1` to data segments in real partitions (10, 20, …), so these common collection-wide deletes are dropped and deleted rows resurface. Fan a `-1` partition group out to all data-segment partitions instead of requiring an exact match.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1735,6 +1803,148 @@ class MilvusScan(
...
+
+    val packedV2Partitions = v2Segments
+      .filter(_.columnGroups.nonEmpty)
+      .map { seg =>
+        MilvusPackedV2InputPartition(
+          segmentID = seg.segmentId,
+          partitionID = seg.partitionId,
+          columnGroups = seg.columnGroups,
+          milvusSchemaBytes = schemaBytes,
+          milvusOption = milvusOption,
```
liliu-z: Driver validation now accepts the `s3.bucketName`/`fs.bucketName` aliases, but the fast path forwards the raw `milvusOption` (`milvusOption = milvusOption`) and the executor-side native StorageV2 reader consumes only the canonical `fs.bucket_name` via `Properties.fromMilvusOption`. If a user sets only an alias with bucket-relative snapshot paths, driver validation passes yet executors fall back to the `a-bucket` default (non-IAM) or throw (IAM) — wrong-bucket or missing-object read failures, and a regression since previously both sides used the same key and rejected it up front. Inject the canonical `fs.bucket_name` into the forwarded options (as `buildClientSnapshotOptions` does, though that helper now has no production caller).

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1798,100 +2001,140 @@ class MilvusScan(
...
+        val pkField = CollectionSchema
+          .parseFrom(schemaBytes)
+          .fields
+          .find(_.isPrimaryKey)
+          .getOrElse {
+            throw new IllegalArgumentException(
+              "No primary key field found in schema"
+            )
+          }
+        MilvusDeltaLogReader.loadPartitionScopedDeletePlans(
```
liliu-z, line 2072: In snapshot mode `planInputPartitionsFromSnapshot` fully decodes every L0 deltalog from S3 (`loadPartitionScopedDeletePlans`), but `buildSnapshotPartitions` uses the result only for a `.contains` boolean and `createReaderFactory` then decodes the identical files again (`:2115`). Since `buildSnapshotHadoopConf` sets `fs.s3a.impl.disable.cache=true`, this is a genuine second network fetch and parse on the driver that scales with delete volume, while plan time only needs `deltaLogs.nonEmpty` per partition. Reuse the already-decoded map (or broadcast it) rather than re-reading — the inherited-L0-deletes fix eliminates this too.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -91,14 +100,23 @@ object V2SegmentLoader extends Logging {
   private[read] def buildV2SegmentInfoFromEntry(
       entry: AvroManifestEntry,
       bucket: String,
-      hadoopConf: Configuration
+      hadoopConf: Configuration,
+      applyDeletes: Boolean = true
   ): Either[Throwable, Option[V2SegmentInfo]] = {
+    val isL0 = entry.segmentLevel == 0L
```
liliu-z: `val isL0 = entry.segmentLevel == 0L` (used at line 115 `else if (isL0 && !applyDeletes)`, logged as "L0 delete-only segment") tests for the wrong enum value. Milvus's SegmentLevel is `Legacy=0, L0=1, L1=2, L2=3`, so this matches Legacy, not L0. Real L0 delete-only segments (level 1) are never recognized, so the applyDeletes=false early-skip never fires for them; conversely any Legacy-level (0) data segment would be silently dropped from an applyDeletes=false (backfill) scan, losing its rows. Compare against `common.SegmentLevel.L0` instead of the literal 0.

PR #103, opened 2026-07-07
fix: [reader] Adapt l0 segment & v3 deltalog

congqixia:
Parse StorageV3 manifest primary-key deltalogs and attach the resulting delete plans to V3 snapshot partitions. Merge inherited L0 delete plans into V3 reads, including collection-wide partition -1 deletes, and apply delete filtering in the Loon reader for normal scans and vector search.

### DIFF
--- a/src/main/scala/read/MilvusLoonPartitionReader.scala
+++ b/src/main/scala/read/MilvusLoonPartitionReader.scala
@@ -39,6 +39,8 @@ import io.milvus.storage.{
 }
 
 object MilvusLoonPartitionReader {
+  private[read] val TimestampColumnName = "1"
+
   private[read] val SystemFieldAliases: Seq[(String, Long)] = Seq(
     "RowID" -> 0L,
     "row_id" -> 0L,
@@ -379,7 +381,9 @@ class MilvusLoonPartitionReader(
         s"StorageV3 delete filtering requires PK column $pkColumnName to be loaded"
       )
     }
-    val rawTsVector = batch.getVector("Timestamp")
+    val rawTsVector = batch.getVector(
+      MilvusLoonPartitionReader.TimestampColumnName
+    )
     if (rawTsVector == null) {
       throw new IllegalStateException(
         "StorageV3 delete filtering requires Timestamp column to be loaded"
@@ -459,7 +463,10 @@ class MilvusLoonPartitionReader(
             "StorageV3 delete filtering requires a primary key field"
           )
         )
-      (requested ++ Seq(pkId, "1")).distinct
+      (requested ++ Seq(
+        pkId,
+        MilvusLoonPartitionReader.TimestampColumnName
+      )).distinct
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/main/scala/read/MilvusLoonPartitionReader.scala
```diff
@@ -353,6 +367,34 @@ class MilvusLoonPartitionReader(
...
+        "StorageV3 delete filtering requires a primary key field"
+      )
+    }
+    val pkVector = batch.getVector(pkColumnName)
+    if (pkVector == null) {
+      throw new IllegalStateException(
+        s"StorageV3 delete filtering requires PK column $pkColumnName to be loaded"
+      )
+    }
+    val rawTsVector = batch.getVector("Timestamp")
```
liliu-z: The Loon reader names its Arrow columns by field-ID string (SchemaUtil sets each field name to `fieldID.toString`, and `getColumnNames` requests the timestamp column as "1"), so `batch.getVector("Timestamp")` at line 382 always resolves to null. It is not dereferenced unchecked and does not throw a NullPointerException: the guard at 383-387 catches the null and throws `IllegalStateException("...requires Timestamp column to be loaded")`. This does not affect every read — only the delete-applying path (`applyDeletes && !deletePlan.isEmpty`, and applyDeletes defaults to true), where it makes every read of a segment that actually carries deletes fail. Resolve the timestamp by its field-ID string "1", the same way `pkColumnName` already resolves the PK.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1853,10 +1879,79 @@ class MilvusScan(
...
+        .find(_.isPrimaryKey)
+      if (pkField.isEmpty) {
+        return Map.empty
+      }
+
+      manifestList.flatMap { item =>
+        val (basePath, readVersion) =
+          MilvusSnapshotReader.parseManifestContent(item.manifest) match {
+            case Right(content) => (content.basePath, content.ver.toLong)
+            case Left(_)        => (item.manifest, -1L)
```
liliu-z, line 1919: On the `Left` branch (manifest not parseable as `{ver, base_path}` JSON) `loadV3DeletePlans` uses readVersion=-1, and the guard at line 1908 (`readVersion <= 0L`) returns None, dropping the segment's own V3 deltalogs. Meanwhile `buildSnapshotPartitions` parses the same manifest identically (1979-1983) and still builds a readable partition at readVersion=-1 (LATEST); the pre-existing `storageV2ManifestBasePath` helper (line 921) shows a raw `item.manifest` is treated as a valid base path. So when a manifest is supplied as a raw path, the segment is read via getLatest but its own deltalog deletes are silently skipped, surfacing deleted rows as live. This is narrow — normal JSON manifests take the Right branch and load deltalogs correctly, and if the raw string is not a readable path the read itself throws rather than leaking. The reviewer's suggested fix is not directly implementable: without a manifest version the `manifest-N.avro` deltalog path cannot be constructed, so it requires latest-version resolution.

PR #104, opened 2026-07-10
feat: [backfill] support vector column backfill

congqixia:
- accept common user input formats for dense and sparse vectors
- normalize vector values to Milvus internal byte layouts before joining
- preserve vector type, dimension, and nullability in Arrow schemas
- validate vector dimensions, ranges, and sparse entries
- add vector normalization and Arrow conversion tests

### DIFF
--- a/src/main/scala/MilvusUtil.scala
+++ b/src/main/scala/MilvusUtil.scala
@@ -542,33 +542,83 @@ object IntConverter {
 }
 
 object SparseFloatVectorConverter {
+  private val MaxIndexExclusive = 0xffffffffL
+
   def encodeSparseFloatVector(sparse: Map[Long, Float]): Array[Byte] = {
-    val sortedEntries = sparse.toSeq.sortBy(_._1)
+    encodeSparseFloatVectorEntries(sparse.toSeq)
+  }
+
+  private[connector] def encodeSparseFloatVectorEntries(
+      entries: Seq[(Long, Float)]
+  ): Array[Byte] = {
+    val sortedEntries = entries.sortBy(_._1)
 
     val buf = ByteBuffer.allocate((4 + 4) * sortedEntries.size)
     buf.order(ByteOrder.LITTLE_ENDIAN)
 
+    var previousIndex: Option[Long] = None
     for ((k, v) <- sortedEntries) {
-      if (k < 0 || k >= Math.pow(2.0, 32) - 1) {
+      if (k < 0 || k >= MaxIndexExclusive) {
         throw new DataParseException(
-          s"Sparse vector index ($k) must be positive and less than 2^32-1"
+          s"Sparse vector index ($k) must be non-negative and less than 2^32-1"
+        )
+      }
+      if (previousIndex.contains(k)) {
+        throw new DataParseException(
+          s"Sparse vector index ($k) is duplicated"
         )
       }
 
-      val lBuf = ByteBuffer.allocate(8)
-      lBuf.order(ByteOrder.LITTLE_ENDIAN)
-      lBuf.putLong(k)
-      buf.put(lBuf.array(), 0, 4)
+      buf.putInt(k.toInt)
 
-      if (v.isNaN || v.isInfinite) {
+      if (v.isNaN || v.isInfinite || v < 0.0f) {
         throw new DataParseException(
-          s"Sparse vector value ($v) cannot be NaN or Infinite"
+          s"Sparse vector value ($v) must be finite and non-negative"
         )
       }
 
       buf.putFloat(v)
+      previousIndex = Some(k)
     }
 
     buf.array()
+  }
+
+  private[connector] def decodeSparseFloatVector(
+      bytes: Array[Byte]
+  ): Seq[(Long, Float)] = {
+    if (bytes.length % 8 != 0) {
+      throw new DataParseException(
+        s"Sparse vector byte length must be a multiple of 8, got ${bytes.length}"
+      )
+    }
+
+    val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
+    val entries = Seq.newBuilder[(Long, Float)]
+    var previousIndex: Option[Long] = None
+    while (buffer.hasRemaining) {
+      val index = buffer.getInt() & 0xffffffffL
+      val value = buffer.getFloat()
+      if (index >= MaxIndexExclusive) {
+        throw new DataParseException(
+          s"Sparse vector index ($index) must be less than 2^32-1"
+        )
+      }
+      previousIndex.foreach { previous =>
+        if (index <= previous) {
+          throw new DataParseException(
+            s"Sparse vector indices must be strictly increasing, got $index after $previous"
+          )
+        }
+      }
+      if (value.isNaN || value.isInfinite || value < 0.0f) {
+        throw new DataParseException(
+          s"Sparse vector value ($value) must be finite and non-negative"
+        )
+      }
+      entries += index -> value
+      previousIndex = Some(index)
+    }
+    entries.result()
   }
 }
--- a/src/main/scala/MilvusUtil.scala
+++ b/src/main/scala/MilvusUtil.scala
@@ -353,34 +353,47 @@ object FloatConverter {
       )
     }
 
-    // IEEE 754 float32 format: 1 sign bit, 8 exponent bits, 23 fraction bits
+    // Match Milvus's float16.Fromfloat32 conversion: IEEE-754
+    // round-to-nearest-even, including subnormal values. Truncating the
+    // mantissa here makes normal ingestion produce different bytes from
+    // Milvus import/backfill for values such as 0.3f and 0.7f.
     val bits = java.lang.Float.floatToRawIntBits(value)
-    val sign = (bits >>> 31) & 0x1
-    val exp = (bits >>> 23) & 0xff
-    val frac = bits & 0x7fffff
+    val sign = (bits >>> 16) & 0x8000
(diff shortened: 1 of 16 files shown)


src/main/scala/serde/SchemaUtil.scala
```diff
@@ -214,18 +241,32 @@ object MilvusSchemaUtil {
...
+        case DataType.BinaryVector | DataType.FloatVector |
+            DataType.Float16Vector | DataType.BFloat16Vector |
+            DataType.Int8Vector =>
+          true
+        case _ => false
+      }
+      val isVector =
+        isDenseVector || field.dataType == DataType.SparseFloatVector
+      val physicalType =
+        if (isDenseVector && field.nullable) new ArrowType.Binary()
```
liliu-z, line 254: This PR emits nullable dense vectors (Float/Float16/BFloat16/Int8) as Arrow `Binary` in the reader's schema (here, plus `convertToArrowSchema` and `convertSparkSchemaToArrow`), but `ArrowConverter.arrowValueToSparkValue` only decodes `FixedSizeBinaryVector`/`ListVector` into the `ArrayType(FloatType)`/`ArrayType(ShortType)` Spark types these fields use, so a `VarBinaryVector` falls through to `vector.asInstanceOf[ListVector]` (ArrowConverter.scala:171/185) and throws `ClassCastException`. Any normal read of a collection containing a nullable dense vector — including backfill's own output — now fails; non-nullable dense vectors stay `FixedSizeBinary` and are unaffected. Fix by adding a `VarBinaryVector` branch to the array-decode cases, or by reading nullable dense vectors as `BinaryType`.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -398,20 +456,78 @@ object ArrowConverter extends Logging {
...
         val floats = (0 until arrayData.numElements())
           .map(i => arrayData.getFloat(i))
           .toArray
-        val bytes = new Array[Byte](floats.length * 4)
-        val buffer = ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN)
-        floats.foreach(buffer.putFloat)
-        vector.asInstanceOf[FixedSizeBinaryVector].set(rowIndex, bytes)
+        val bytes = milvusType match {
+          case Some(MilvusDataType.Float16Vector) =>
+            floats.flatMap(value => FloatConverter.toFloat16Bytes(value))
```
liliu-z, line 629: The general write path here (and normal ingestion at MilvusUtil.scala:272) encode Float16 via FloatConverter.toFloat16Bytes, which truncates the mantissa (frac >>> 13), while the backfill normalizer encodes via VectorBackfillSupport.floatToHalfBits, which rounds to nearest-even. The same float32 therefore produces different 2-byte Float16 values depending on which path wrote it (e.g. 0.3 -> 0x34cc truncated vs 0x34cd rounded; 0.7 also differs), so a vector backfilled here is not byte-identical to the same vector written through normal ingestion. Route both paths through one Float16 encoder.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -398,20 +456,78 @@ object ArrowConverter extends Logging {
           vector.asInstanceOf[VarCharVector].setSafe(rowIndex, str.getBytes)
         }
 
-      case ArrayType(FloatType, _) =>
-        // FloatVector stored as FixedSizeBinaryVector
+      case ArrayType(FloatType, _) if isBinaryBackedVector(vector) =>
```
liliu-z, line 616: This ArrayType(FloatType) branch builds `bytes` from arrayData.numElements() with no comparison to the field's expected dimension, then writes via setBinaryVectorValue, which for nullable dense vectors calls VarBinaryVector.setSafe (accepts any length). The BinaryType and Array[Short]/Array[Byte] branches share the same unguarded setBinaryVectorValue path. So a wrong-dimension input on the general (non-backfill) nullable dense-vector path is silently stored as a malformed vector. The non-nullable FixedSizeBinaryVector.set path fails fast on width mismatch, and backfill is separately guarded by requireDimension/requireByteWidth in VectorBackfillSupport, so only this raw nullable path is exposed. Validate encoded width against the dimension metadata (MilvusVectorDimensionMetadataKey) before setSafe.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -442,8 +665,14 @@ object ArrowConverter extends Logging {
...
+            vector,
+            rowIndex,
+            bytes,
+            "BinaryType",
+            milvusType,
+            vectorDimension
+          )
         }
 
       case MapType(keyType, valueType, _) =>
```
liliu-z, line 791: This PR's convertSparkSchemaToArrow now maps any field carrying SparseFloatVector metadata to ArrowType.Binary (SchemaUtil.scala:374), so a reader-produced sparse field — emitted as MapType(LongType,FloatType) with milvus.data_type=104 and no dim (MilvusSnapshotReader.scala:869,821) — becomes a VarBinaryVector at runtime. But the internalRowToArrow write branch still matches Spark MapType and does vector.asInstanceOf[MapVector] (line 679), throwing a deterministic ClassCastException on every sparse row. Both writers that pair these functions (MilvusLoonPartitionWriter:232/388, MilvusV2BinlogWriter:131/203) are affected, so any Spark read-then-write of a collection with a sparse vector fails immediately. The backfill path itself is safe because it pre-normalizes sparse fields to BinaryType via canonicalStructField. Encode MapType into the sparse byte layout in internalRowToArrow so the write path matches the Binary vector the schema now produces.

src/main/scala/serde/SchemaUtil.scala
```diff
@@ -277,48 +318,109 @@ object MilvusSchemaUtil {
...
+          .filter(
+            _.contains(ArrowConverter.MilvusVectorDimensionMetadataKey)
+          )
+          .map(
+            _.getLong(ArrowConverter.MilvusVectorDimensionMetadataKey).toInt
+          )
+          .orElse(vectorDimensions.get(field.name))
+          .filter(_ > 0)
+          .getOrElse {
+            throw new IllegalArgumentException(
```
liliu-z, line 358: The PR's rewritten convertSparkSchemaToArrow treats any field whose metadata carries a vector milvus.data_type as a vector and eagerly resolves its dimension (SchemaUtil.scala:366), throwing when none is found (line 358). Reader-produced schemas set milvus.data_type but never the dimension metadata (MilvusSnapshotReader.fieldToStructField), and MilvusLoonWriter.extractVectorDimensions only collects dims for Array[Float] columns, so a BinaryVector (Spark BinaryType) or Int8Vector (Spark ArrayType(ShortType)) read from Milvus has no dimension source and hits this throw at Arrow-schema build time. BinaryVector previously wrote through as raw Binary without a dimension, so this is a behavioral regression on the Spark read-then-write path; Int8Vector has no options-based workaround because extractVectorDimensions ignores non-float element types entirely. Note the throw is in convertSparkSchemaToArrow, not extractVectorDimensions, and Float16/BFloat16 are mapped to Array[Float] by the reader so they are collected. Propagate the vector dimension from the reader (emit MilvusVectorDimensionMetadataKey), or extend extractVectorDimensions to non-float vector types.

src/main/scala/MilvusUtil.scala
```diff
@@ -353,34 +353,47 @@ object FloatConverter {
...
-      val newExp = exp - 127 + 15
-      if (newExp < 0) 0
-      else if (newExp > 0x1f) 0x1f
-      else newExp
+      var half = sign | (halfExponent << 10) | (mantissa >>> 13)
+      val remainder = mantissa & 0x1fff
+      if (remainder > 0x1000 || (remainder == 0x1000 && (half & 1) != 0)) {
+        half += 1
+      }
+      if (((half >>> 10) & 0x1f) == 0x1f) {
```
liliu-z, line 393: toFloat16Bytes rejects magnitudes in (65504, 65520) — e.g. 65510f — via the pre-existing isWithinFloat16Range guard (absValue <= 65504f) before the new round-to-nearest logic runs, whereas Milvus's float16 conversion rounds that band down to the max normal 0x7bff (65504) and only >= 65520 overflows to infinity. So a float16 backfill aborts on values Milvus would accept, defeating the rewrite's stated goal of matching Milvus; the new overflow throw at line 390 is also dead, since nothing that clears the <=65504 gate can round up to 0x1f. Relax the guard to reject only magnitudes >= 65520 and let the rounding path handle the boundary.

src/main/scala/serde/ArrowConverter.scala
```diff
@@ -398,20 +600,99 @@ object ArrowConverter extends Logging {
...
+          bytes,
+          "Array[Float]",
+          milvusType,
+          vectorDimension
+        )
+
+      case ArrayType(ShortType, _) if isBinaryBackedVector(vector) =>
+        val arrayData = record.getArray(colIndex)
+        val values = (0 until arrayData.numElements())
+          .map(i => arrayData.getShort(i))
```
liliu-z: The new binary-backed vector encode paths read array elements with null-blind accessors — getFloat (606), getShort (632), getByte (665) — so a null element inside a vector array is silently encoded as 0 instead of being rejected, unlike the sibling generic-list branch (isNullAt guard at 704) and the sparse-map branch (isNullAt at 343). This affects only the general Array-typed write path; the backfill feature itself is not exposed, because VectorBackfillSupport.normalizeVectorColumns rejects null vector elements via requireNoNullElements and emits BinaryType, which takes the BinaryType branch rather than these ArrayType branches. Add an isNullAt(i) check before each element read.

liliu-z (commented):
/lgtm
/approve

PR #107, opened 2026-08-05
enhance: [backfill] decouple join pipeline from primary keys

congqixia:
- model resolved join keys as component sequences with internal aliases
- separate backfill join columns from fields written to storage
- generalize source projection, key validation, and merge joins
- preserve the existing PK-only configuration and result JSON contract
- add composite-key seam tests and document the PR1/PR2 split

### DIFF
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -141,7 +141,7 @@ object MilvusBackfill {
       // Step 3: Resolve the current primary-key join into the generic runtime
       // join-key model. PR1 intentionally keeps PK as the only public behavior;
       // PR2 will add explicit physical-field resolution on top of this seam.
-      val joinKey = snapshotMetadataOpt match {
+      val baseJoinKey = snapshotMetadataOpt match {
         case Some(metadata) =>
           val pkField = metadata.collection.schema.fields
             .find(_.isPrimaryKey.getOrElse(false))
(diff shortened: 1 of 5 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -583,19 +574,147 @@ object MilvusBackfill {
...
+      val missingKeys = joinKey.sourceColumns.filterNot(mappedColumns.contains)
+      if (missingKeys.nonEmpty) {
+        Left(
+          SchemaValidationError(
+            s"Backfill data is missing join-key columns after column mapping: " +
+              missingKeys.mkString(", ")
+          )
+        )
+      } else {
+        val reservedCollisions = joinKey.internalColumns.filter(name =>
```
liliu-z: The join key is renamed to a hardcoded internal alias `__bf_join_0__` (BackfillJoinKey.scala:35-36). A non-key target field named exactly `__bf_join_0__` is rejected by this `reservedCollisions` guard with a message telling the user to rename it or use `--column-mapping`, but neither helps: the target must keep that exact name to resolve through the `fieldNameToId` lookup at lines 226/235, so such a field can never be backfilled. The guard also uses case-sensitive `Seq.contains` while Spark's default resolver is case-insensitive, so a case variant like `__BF_JOIN_0__` slips through and later fails with an opaque `[AMBIGUOUS_REFERENCE]`. Both cases fail loudly on a very unlikely field name (no corruption), but a collection PK named `__bf_join_0__` does work, so the fix is narrow: derive the alias so it cannot equal any mapped/collection column under the configured resolver.

congqixia: Agreed. Milvus allows __bf_join_0__ as a user field and resolves field names case-sensitively, while Spark is case-insensitive by default.
We’ll allocate the internal join alias against the mapped and collection columns using Spark’s configured resolver, carry the resolved alias through ResolvedJoinKey, and add exact/case-variant collision tests.

liliu-z (commented):
/lgtm
/approve

PR #108, opened 2026-08-08
fix: [backfill] honor platform AssumeRole configuration

tsunghanjacktsai:
## Summary

- Preserve platform-injected Hadoop S3A `AssumedRoleCredentialProvider` settings, including global Data Role and per-bucket External Volume roles.
- Preserve platform-injected Alibaba OSS AssumeRole settings and forward the global main-storage role from `fs.oss.assumed.role.*` to milvus-storage Native.
- Forward the global main-storage role to milvus-storage Native through `fs.role_arn`, `fs.session_name`, and `fs.external_id`.
- Add `--s3-cloud-provider` so Backfill can select the native storage provider (`aws` by default, `aliyun` for OSS).
- Fail fast when AWS S3A or Alibaba OSS role-provider configuration is incomplete, instead of silently falling back to the pod identity.
- Keep existing static AK/SK and plain IAM/IRSA behavior unchanged.

## Design

The platform remains responsible for selecting and injecting roles. Backfill consumes the existing Hadoop configuration and mirrors the global main-storage role to Native storage:

- AWS: `fs.s3a.assumed.role.*` -> `fs.role_arn`, `fs.session_name`, `fs.external_id`
- Alibaba Cloud: `fs.oss.assumed.role.*` -> `fs.role_arn`, `fs.session_name`, `fs.external_id`

Only the global role is forwarded to Native because Native reads and writes Milvus main storage. Per-bucket customer roles remain scoped to Hadoop reads of the corresponding input bucket.

## Validation

- `sbt scalafmtCheckAll`
- `sbt "testOnly com.zilliz.spark.connector.operations.backfill.BackfillConfigTest com.zilliz.spark.connector.operations.backfill.BackfillAppTest"`
- `git diff --check`

### DIFF
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -108,17 +108,13 @@ object BackfillApp {
     spark.sparkContext.setLogLevel("WARN")
 
     try {
-      val config = baseConfig.withHadoopStorageAssumeRole(
-        spark.sparkContext.hadoopConfiguration,
-        spark.sparkContext.applicationId
-      )
-      MilvusBackfill.run(spark, parquetPath, snapshotPath, config) match {
+      MilvusBackfill.run(spark, parquetPath, snapshotPath, baseConfig) match {
         case Right(result) =>
           println(result.summary)
           println(result.segmentSummary)
           parsed.get("output-result").foreach { outputPath =>
             MilvusBackfill
-              .writeResultJson(spark, result, outputPath, config) match {
+              .writeResultJson(spark, result, outputPath, baseConfig) match {
               case Right(_) =>
                 println(s"Result JSON written to: $outputPath")
               case Left(err) =>
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -108,7 +108,7 @@ object BackfillApp {
     spark.sparkContext.setLogLevel("WARN")
 
     try {
-      val config = baseConfig.withHadoopS3AssumeRole(
+      val config = baseConfig.withHadoopStorageAssumeRole(
         spark.sparkContext.hadoopConfiguration,
         spark.sparkContext.applicationId
       )
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -11,7 +11,8 @@ import com.zilliz.spark.connector.MilvusOption
   * spark-connector-assembly.jar \ --parquet <path> --snapshot <path> \
   * --s3-endpoint <endpoint> --s3-bucket <bucket> \ --s3-access-key <key>
   * --s3-secret-key <secret> \ [--s3-root-path <path>] [--s3-region <region>]
-  * [--s3-use-ssl] \ [--batch-size <n>] [--output-result <path>] \ [--mode
+  * [--s3-cloud-provider aws|aliyun|gcp|azure|tencent|huawei] [--s3-use-ssl] \
+  * [--batch-size <n>] [--output-result <path>] \ [--mode
   * replace|coalesce|overwrite]
   *
   * --mode:
@@ -70,6 +71,10 @@ object BackfillApp {
       s3UseSSL = parsed.contains("s3-use-ssl"),
       s3RootPath = parsed.getOrElse("s3-root-path", "files"),
       s3Region = parsed.getOrElse("s3-region", "us-east-1"),
+      s3CloudProvider = parsed.getOrElse(
+        "s3-cloud-provider",
+        BackfillConfig.DefaultCloudProvider
+      ),
       s3UseIam = useIam,
       sourceS3Endpoint = parsed.get("source-s3-endpoint"),
       sourceS3AccessKey = parsed.get("source-s3-access-key"),
@@ -153,6 +158,7 @@ object BackfillApp {
     "s3-secret-key",
     "s3-root-path",
     "s3-region",
+    "s3-cloud-provider",
     "source-s3-endpoint",
     "source-s3-access-key",
     "source-s3-secret-key",
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -62,7 +62,7 @@ object BackfillApp {
       ) Some(false)
       else None
 
-    val config = BackfillConfig(
+    val baseConfig = BackfillConfig(
       s3Endpoint = s3Endpoint,
       s3BucketName = s3Bucket,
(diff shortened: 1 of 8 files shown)


src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -103,6 +108,10 @@ object BackfillApp {
     spark.sparkContext.setLogLevel("WARN")
 
     try {
+      val config = baseConfig.withHadoopStorageAssumeRole(
```
liliu-z: The AssumeRole derivation is wired only into the CLI entry point, and `withHadoopStorageAssumeRole` is `private[backfill]`, so an embedded caller of the documented `MilvusBackfill.run` API cannot invoke it and has no documented alternative (the README config reference never mentions `s3RoleArn`/`s3RoleSessionName`/`s3ExternalId`). Such a caller running on a platform that injects a global S3A assumed role gets an asymmetric job: `configureHadoopS3ForPath` preserves the assumed-role provider for the Hadoop reads, while the native milvus-storage writer receives no `fs.role_arn` and signs with the pod's ambient identity. `run()` already reads and mutates the session's Hadoop configuration, so moving the derivation inside it costs no purity that has not already been spent.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1829,6 +1829,40 @@ object MilvusBackfill {
...
+          hadoopConf.getTrimmed(s"$prefix.assumed.role.arn")
+        ).forall(_.isEmpty)
+      ) {
+        throw new IllegalArgumentException(
+          s"$prefix.assumed.role.arn must be set when $bucketProviderKey uses " +
+            BackfillConfig.HadoopS3AssumedRoleProvider
+        )
+      }
+      if (
+        !bucketUsesAssumeRole && globalUsesAssumeRole && Option(
```
liliu-z: The check `!bucketUsesAssumeRole && globalUsesAssumeRole && Option(hadoopConf.getTrimmed(BackfillConfig.HadoopS3AssumedRoleArn)).forall(_.isEmpty)` inspects only the global ARN key, but Hadoop S3A resolves `fs.s3a.bucket.<bucket>.*` over the global setting, so provider and ARN can legitimately come from different scopes. A user who sets the credential provider globally and the ARN at `fs.s3a.bucket.<bucket>.assumed.role.arn` (or the inverse, which the preceding check rejects) has a fully valid S3A setup, yet backfill aborts with a configuration error before reading a single row. Resolve the effective provider and ARN each as bucket value with global fallback, then validate that resolved pair.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -251,10 +269,178 @@ case class BackfillConfig(
...
+  }
+
+  private def withAwsS3AssumeRole(
+      hadoopConf: Configuration,
+      defaultSessionName: String
+  ): BackfillConfig = {
+    val provider = Option(
+      hadoopConf.getTrimmed(BackfillConfig.HadoopS3CredentialsProvider)
+    ).getOrElse("")
+    if (!BackfillConfig.isAssumedRoleProvider(provider)) {
```
liliu-z: The native-side derivation only looks at the global `fs.s3a.aws.credentials.provider` / `fs.s3a.assumed.role.arn` keys, so when the platform scopes the main-storage role per bucket (`fs.s3a.bucket.<milvus-bucket>.assumed.role.*`) it returns the config unchanged and no `fs.role_arn` reaches milvus-storage. `configureHadoopS3ForPath` explicitly supports that bucket-scoped case for the Hadoop side, so the job ends up reading through the customer role while the native writer silently falls back to the pod identity — the exact silent fallback this PR sets out to eliminate. Either resolve the main bucket's role with bucket-scoped keys taking precedence over global ones, or fail loudly when a bucket-scoped role is present but not forwarded.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -251,10 +269,178 @@ case class BackfillConfig(
...
+  private[backfill] def withHadoopStorageAssumeRole(
+      hadoopConf: Configuration,
+      defaultSessionName: String
+  ): BackfillConfig = {
+    if (!s3UseIam || s3RoleArn.exists(_.trim.nonEmpty)) {
+      this
+    } else if (s3CloudProvider.trim == "aliyun") {
+      withAlibabaOssAssumeRole(hadoopConf, defaultSessionName)
+    } else {
+      withAwsS3AssumeRole(hadoopConf, defaultSessionName)
```
liliu-z: Every cloud provider except `aliyun` is routed through the AWS S3A derivation, so a `--s3-cloud-provider tencent` or `huawei` job whose Hadoop configuration carries a global `fs.s3a` assumed-role setup gets `fs.role_arn` forwarded to milvus-storage, which rejects it outright for those providers instead of using its normal STS chain. Gate the derivation on `s3CloudProvider` being `aws` rather than treating non-aliyun as AWS.

tsunghanjacktsai:
@liliu-z I've corrected the implementation according to your comments. Please review it again, thx.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -103,10 +116,23 @@ case class BackfillConfig(
...
+      s3RoleSessionName.exists(_.trim.nonEmpty) || s3ExternalId.exists(
+        _.trim.nonEmpty
+      )
+
     if (s3Endpoint.isEmpty) {
       Left("s3Endpoint cannot be empty")
     } else if (s3BucketName.isEmpty) {
       Left("s3BucketName cannot be empty")
+    } else if (
+      !BackfillConfig.AllowedCloudProviders.contains(s3CloudProvider.trim)
```
congqixia: validate() currently accepts s3RoleArn for every allowed cloud provider, and withS3Authentication() then forwards it to Native storage. However, milvus-storage only supports role_arn for AWS and Aliyun; providers such as GCP, Tencent, Huawei, and Azure will fail later during filesystem initialization. This affects embedded callers that construct BackfillConfig directly and conflicts with the PR’s fail-fast goal. Please reject s3RoleArn unless s3CloudProvider is aws or aliyun.

congqixia (approved):
/lgtm

src/main/scala/operations/backfill/BackfillConfig.scala
tsunghanjacktsai: Fixed in 1a1baf1. BackfillConfig.validate now rejects s3RoleArn unless s3CloudProvider is aws or aliyun, matching the Native storage providers that support role_arn. Added coverage for accepted aws/aliyun roleArn configs and fail-fast rejection for gcp/azure/tencent/huawei.

liliu-z (commented):
/lgtm
/approve

PR #109, opened 2026-08-11
feat: Support physical join keys for backfill

congqixia:
Keep primary-key joins as the default while allowing callers to select a persisted scalar snapshot field through the API or `--join-key`.

Resolve configured keys before source reads, normalize them to internal aliases, and validate exact types. Cardinality is enforced on the backfill-input side: the parquet join key must be non-null and unique so a physical source row never fans out. Source-side values may repeat (many-to-one); the same parquet record is applied to every matching source row. The physical key field itself must be declared non-nullable in the snapshot schema. Keep identity fields out of writer targets so result JSON and per-segment write behavior remain backward compatible.

Reserve every internal column the join/merge pipeline synthesizes (match flag, join aliases, per-field provenance flags, backfill-side suffix) under Spark's configured resolver before any source read, so a target such as `__BF_MATCHED__` can no longer be overwritten by marker data.

Document mapping and validation constraints (README, user guide, checked-in plan) and add regression coverage for configuration, resolution, source projection, merge modes, internal-name reservation, and the physical-key-requires-snapshot guard in `run()`.


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -359,15 +475,40 @@ object MilvusBackfill {
         case Right(df)   => df
       }
 
+      val reusableOriginalDF = config.joinKey match {
+        case _: BackfillJoinKey.PhysicalField =>
+          originalDF.persist(StorageLevel.MEMORY_AND_DISK)
+          cachedSourceDF = originalDF
+          validateJoinKeyCardinality(
```
liliu-z: The new source-side join-key check runs against a read that deliberately keeps tombstones (`ReadApplyDeletes -> "false"` at :980), so a deleted row and its live replacement sharing the same key are counted as duplicates and the backfill aborts even though the live data is unique; any collection that has been upserted or deleted on the join field fails here, including the `--join-key <pk_name>` spelling of the default key. The fix is delete-aware validation rather than dropping the check, since an arbitrary physical field has no server-side uniqueness guarantee and two genuinely live rows with the same key would both be rewritten under overwrite/replace. A delete-applied read projecting only the join column keeps that cost well below a second full scan, because the target fields do not need re-reading.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -45,6 +55,106 @@ object MilvusBackfill {
...
   private[backfill] val RowOffsetCol = MilvusOption.MilvusExtraColumnRowOffset
 
+  private val SupportedPhysicalJoinSparkTypes
+      : Set[org.apache.spark.sql.types.DataType] = Set(
+    ByteType,
+    ShortType,
+    IntegerType,
+    LongType,
+    StringType,
+    BinaryType
```
liliu-z: `SupportedPhysicalJoinSparkTypes` allowlists `BinaryType`, but no mapped Milvus scalar type produces `BinaryType`: `dataTypeToSparkType` yields it only for `BinaryVector` (already rejected as a vector) and for its `case _` unknown-type fallback, so the entry's sole effect is to admit fields whose Milvus type this connector cannot interpret — including any future vector type, which `isVectorField` also would not recognize. `BackfillModeTest.scala:208` pins `27 -> BinaryType` as a supported key type, so gating on a known scalar Milvus type code instead of the derived Spark type requires updating that case as well.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -359,15 +475,40 @@ object MilvusBackfill {
         case Right(df)   => df
       }
 
+      val reusableOriginalDF = config.joinKey match {
+        case _: BackfillJoinKey.PhysicalField =>
+          originalDF.persist(StorageLevel.MEMORY_AND_DISK)
```
liliu-z: `validateJoinKeyCompatibility` is a schema-only check that needs nothing but `originalDF.schema`, yet it runs after the source DataFrame has been persisted and fully scanned by `validateJoinKeyCardinality`. A join-key type mismatch between the parquet and the snapshot therefore costs a full read and cache of the entire collection before it is reported; hoisting the compatibility check above the persist makes that case fail in milliseconds.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -592,25 +692,111 @@ object MilvusBackfill {
...
+    if (!mapping.values.toSet.contains(joinFieldName)) {
+      return Left(
+        SchemaValidationError(
+          s"column mapping must include physical join-key field " +
+            s"'$joinFieldName' as a target"
+        )
+      )
+    }
+
+    val targetFields = mapping.values.toSet - joinFieldName
```
liliu-z, line 797: In the physical-join-key path, `targetFields` excludes only the join field, so nothing stops the collection's primary key from becoming a writer target. The legacy path can never hit this because it requires the PK as the mapping target and then subtracts it (line 676), and it even guards the implicit-rename collision at line 631. The physical path has no counterpart: when `--column-mapping` is omitted, `mapping` is an identity map over *every* parquet column (line 716), so a parquet exported with the PK column alongside the join key silently makes the PK a backfill target. Downstream validation does not catch it — `newFieldNames` is only checked for presence in the snapshot schema (line 339) and `grep isPrimaryKey/isPartitionKey` over `src/main/scala/operations/backfill/` finds no guard anywhere in the write path. In coalesce/overwrite the PK field ID is then added to `extraReadFields` and re-written as a new binlog/column group; in `replace` the join is a plain left join (line 1310), so unmatched rows get a NULL PK. This also contradicts `docs/user-guide-snapshot-backfill.md:25`, which lists primary-key columns as unsupported. Validate targets against the collection schema (reject `isPrimaryKey`/`isPartitionKey`/system fields) in a shared helper both paths call, rather than duplicating the exclusion in two places that have already diverged.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -44,6 +46,94 @@ object MilvusBackfill {
...
+              s"Physical join-key field '$name' was not found in snapshot schema. " +
+                s"Available fields: ${schema.fields.map(_.name).mkString(", ")}"
+            )
+          )
+    }
+
+    selected.flatMap { field =>
+      spec match {
+        case _: BackfillJoinKey.PhysicalField
+            if !SupportedPhysicalJoinMilvusTypes.contains(field.dataType) =>
```
liliu-z, line 124: `resolveJoinKey` validates only the Milvus type code of the physical join field and never inspects `Field.nullable`, and commit 5ac5905 removed the `validateJoinKeyCardinality(originalDF, ...)` call that used to reject null source keys — the remaining call at line 388 covers the parquet side only. So `--join-key some_nullable_field` is accepted, and every source row whose key is null is permanently unmatchable (Spark equi-joins never match nulls). In coalesce/overwrite that just means those rows are silently skipped; in `replace` the left join leaves the target columns NULL for them, with no count or warning distinguishing "no parquet row for this key" from "key was null". Either reject a nullable field at resolve time or report the null-key source row count so the user can see the coverage gap.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,446 @@
...
+    collection_name="my_collection",
+    field_name="score",
+    data_type=DataType.FLOAT,
+)
+```
+
+After this call, `describe_collection` will show the new fields, but
+existing rows have no values for them.
+
+> **Tip.** Do this **after** `CreateSnapshot`, not before — otherwise the
```
liliu-z: The tip tells users to snapshot first and add the field afterwards, but the backfill resolves target field IDs from the snapshot's own schema and aborts if a target is absent, so a field added after the snapshot can never be backfilled against it. The §3 flow diagram at lines 45-46 shows the same inverted order, so both places need fixing: add the field first, then take the snapshot, and state the schema-evolution constraint explicitly.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,446 @@
...
+from source side).
+
+**`coalesce` / `overwrite` caveats** (shared — both read source-side values):
+
+- Parquet types must match the collection field types **exactly**
+  (e.g. Int64 vs Int32 will be rejected). Spark's type widening would
+  silently change the stored Arrow type otherwise.
+- Reads the target field(s) from the base segments — slightly heavier I/O
+  than `replace`.
+- For `coalesce`: if the target field has no prior binlogs (you just added
```
liliu-z: The caveat says coalesce onto a just-added field "reads return all nulls" and is "Safe", but on V2 packed segments the reader throws rather than returning nulls: coalesce puts every target field ID into `ReaderFieldIDs`, and the packed reader rejects any requested field ID that no existing column group declares — exactly the case for a field with no prior data. Mark this as unsupported on V2 and describe the packed-reader restriction.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,446 @@
...
+### Required
+
+| Flag              | Description                                                             |
+| ----------------- | ----------------------------------------------------------------------- |
+| `--parquet`       | Path to user Parquet with join-key + new-field columns. `s3a://` or local. |
+| `--snapshot`      | Path to `snapshot.json` produced by Milvus `CreateSnapshot`.            |
+| `--s3-endpoint`   | S3 endpoint for Milvus storage (where segments live).                   |
+| `--s3-bucket`     | Bucket for Milvus storage.                                              |
+
+Omitting `--snapshot` switches the job to **client mode**, which requires
```
liliu-z: The client-mode prose here and the flag table at lines 376-384 document `--milvus-uri`, `--milvus-token`, `--database`, `--collection`, and `--partition`, but none of those exist in `KvFlags` or `BoolFlags`, so `parseArgs` throws `Unknown argument` on the first one it sees. Anyone following this section gets a hard failure before the job starts. Either drop the section or rewrite it against the flags the parser actually accepts.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,446 @@
...
+  fill in values for rows that already exist.
+- Re-compute a scalar/text/JSON field from an external system (e.g. an
+  embedding classifier output, a new metadata column, a re-scored relevance
+  label) without re-ingesting base data.
+- Do the above without load on the Milvus cluster and without rebuilding
+  indexes for existing fields.
+
+**Not supported today:**
+
+- Backfilling vector fields, primary-key columns, dynamic fields, or
```
liliu-z: This line lists vector backfill as unsupported while the backfill README in the same PR documents six vector types with their accepted input encodings, so readers get contradictory guidance on a headline capability. Separately, none of the seven files linked in §11 (lines 440-446) exist in the repo, and the same missing docs are referenced again at lines 9, 394, and 425, as is an `examples/` directory that isn't present (lines 193, 198). Align the vector statement with the README and either add the referenced docs or drop the links.

src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -162,11 +112,71 @@ object BackfillApp {
...
+      sourceS3UseSSL =
+        if (parsed.contains("source-s3-use-ssl")) Some(true) else None,
+      sourceS3UseIam = sourceUseIamFlag,
+      sourceS3Region = parsed.get("source-s3-region"),
+      batchSize = parsed.getOrElse("batch-size", "1024").toInt,
+      columnMapping = parsed.get("column-mapping").map(parseColumnMapping),
+      mode = parsed.getOrElse("mode", MilvusOption.BackfillModeCoalesce),
+      joinKey = parsed
+        .get("join-key")
+        .map(value => BackfillJoinKey.PhysicalField(value.trim))
```
liliu-z, line 175: The CLI path trims the `--join-key` value while callers of the programmatic API pass it through untrimmed, so a value with stray whitespace reaches lookup unchanged. The resulting error reports the field as not found while listing that very field among the available ones, which reads as a contradiction and sends users hunting for a schema problem that does not exist. Trim in the shared validation path instead of only in `parseArgs`.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,422 @@
...
+- `describe_collection` → confirm schema includes new fields.
+- `query(expr="pk in [...]", output_fields=["category", "score"])` → confirm values.
+- `get_segment_info` → confirm every segment's `DataVersion` (or
+  `manifest_version`) moved forward.
+
+## 6. Modes: replace vs coalesce vs overwrite
+
+Pass `--mode coalesce` (default), `--mode overwrite`, or `--mode replace`.
+
+All three behave identically for matched rows that already have non-null
```
liliu-z: This sentence says all three modes behave identically for matched rows whose target value is non-null, but `performJoin` does the opposite for exactly that case: coalesce keeps the source value (`coalesce(df.col(n), df.col(n + suffix))` at MilvusBackfill.scala:1316 → "old"), overwrite takes the parquet value (`when(matched, df.col(n + suffix))` at :1340 → "new"), and replace takes the parquet column verbatim. The comparison table three lines below is correct, but this paragraph is what a user reads first when choosing a mode, so it will steer them into silently keeping or clobbering values they did not intend; rewrite the sentence to agree with the table.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -44,6 +47,159 @@ object MilvusBackfill {
...
+              "No primary key field found in snapshot schema"
+            )
+          )
+      case BackfillJoinKey.PhysicalField(name) if name.isEmpty =>
+        Left(
+          SchemaValidationError("Physical join-key field name cannot be blank")
+        )
+      case BackfillJoinKey.PhysicalField(name) =>
+        schema.fields
+          .find(_.name == name)
```
liliu-z, line 102: `resolveJoinKey` accepts any snapshot field name for `--join-key`, including the reserved metadata aliases `segment_id` and `row_offset`. Backfill always requests the `$segment_id` / `$row_offset` extra columns, and `MilvusDataSource.appendExtraColumns` rejects a supplied schema containing either alias with `IllegalArgumentException("Field 'segment_id' is a legacy alias for metadata extra column '$segment_id'; use '$segment_id' in the schema or remove it...")`. A user whose collection genuinely has a field named `segment_id` gets that message from deep inside the reader with no achievable remedy — you cannot rename a persisted Milvus field. Reject those two names in `resolveJoinKey` with an error that names `--join-key` instead.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,424 @@
...
+- **V3** — `version` + `manifestPaths`.
+- **V2** — `storage_version: 2` + `column_groups[]`.
+
+### 5.5 Commit to Milvus
+
+Iterate the `segments` map and, for each entry, call
+`UpdateSegmentColumnGroups` on Milvus. The shape depends on V2 vs V3:
+
+```python
+# Pseudocode — use your SDK's equivalent
```
liliu-z: Step 5.5 tells users to iterate the `segments` map and call `client.update_segment_column_groups(...)` / `client.update_segment_manifest(...)`, described as "use your SDK's equivalent". No such SDK methods exist — a code search of milvus-io/pymilvus returns zero hits for `update_segment_column_groups`, and `UpdateSegmentColumnGroups` appears only in Milvus's `internal/datacoord/*.go`, never in a proto service. The actual commit path is a single HTTP GET to `/management/datacoord/backfill/commit?result_path=<backfill_result.json>`, which the proxy handler `CommitBackfillResult` serves. Replace the per-segment SDK pseudocode with that call; line 41's claim that "Milvus gRPC is needed for ... result commit" and the line 403 troubleshooting row need the same correction.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,424 @@
...
+- Using vector fields as physical join keys. Supported vector fields can still
+  be backfill targets.
+- Collections whose sealed segments are still in the legacy V1 binlog
+  format. You need a cluster that writes storage-V2 or V3 segments.
+
+## 2. What you need
+
+| Requirement                | Notes                                                                                           |
+| -------------------------- | ----------------------------------------------------------------------------------------------- |
+| Milvus server              | Snapshot + `UpdateSegmentColumnGroups` support required (any recent 2.6+ or internal build).    |
```
liliu-z: The requirements table states the server needs "Snapshot + `UpdateSegmentColumnGroups` support ... (any recent 2.6+ or internal build)", but the backfill commit route does not exist in any 2.6 release — `internal/http/router.go` has zero backfill references at v2.6.0, v2.6.5, and v2.6.10, and gains `RouteCommitBackfill` only at v3.0.0. A 2.6 user will run the whole Spark job and then hit a 404 at the commit step with nothing pointing at their server version. State v3.0.0 as the floor for the commit path.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+reserved for backfill metadata and cannot be used as join keys.
+`$row_offset` only restores segment write order and is not a stable row identity.
+
+**Type rules:**
+
+- Join-key type in Parquet must match the selected snapshot field type exactly
+  (int64 ↔ Int64, varchar ↔ String).
+- In `--mode coalesce` (default) or `--mode overwrite`, Parquet types must
+  match the Milvus field types **exactly** — no widening. See "Modes" below.
+- In `--mode replace`, Parquet types must be compatible with the Milvus
```
liliu-z: The guide tells users that `replace` mode casts Parquet values to the target field type, and line 391 recommends switching to `replace` "if you're OK with Spark widening" as the fix for a type-mismatch error. `replace` performs no cast: `performJoin` returns a bare left join, `validateMergeableFieldTypes` is skipped because it is gated on `readsSourceFields` (false for `replace`), and the writer schema is taken straight from the joined DataFrame's Parquet types. A user who follows the line-391 advice writes binlogs whose Arrow type does not match the Milvus field, with no error at any stage. Either document that `replace` requires the caller to pre-cast, or add the cast. Note `src/main/scala/operations/backfill/README.md:176` carries the same claim ("Only `replace` lets Spark cast between compatible numeric widths") and should be fixed together.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+- Collections whose sealed segments are still in the legacy V1 binlog
+  format. You need a cluster that writes storage-V2 or V3 segments.
+
+## 2. What you need
+
+| Requirement                | Notes                                                                                           |
+| -------------------------- | ----------------------------------------------------------------------------------------------- |
+| Milvus server              | Milvus 3.0.0+ with snapshot support and the backfill commit management endpoint.                |
+| Object storage             | S3 / MinIO / GCS with S3-compatible endpoint. Must be accessible from both Milvus and Spark.    |
+| Spark                      | 3.5.x, JDK 8+. Cluster mode on YARN, Kubernetes (Spark Operator), or standalone all work.       |
```
liliu-z: The prerequisites table states Spark 3.5.x and JDK 8+, which no other part of the project agrees with: the build pins Spark 4.0.0, the README requires Java 21 and Spark 4.0.1, and `MilvusClient.scala` imports `java.net.http`, which does not exist before JDK 11. Someone provisioning from this table gets an environment the connector will not run on. (The reviewer cited build.sbt:34 for the Spark version; that line is the Scala version — the Spark pin is in project/Dependencies.scala:6.)

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+## 5. Step-by-step
+
+### 5.1 Add the new field
+
+```python
+from pymilvus import DataType, MilvusClient
+
+client = MilvusClient(uri="http://milvus:19530")
+
+client.add_collection_field(
```
liliu-z, line 130: Both `add_collection_field` examples (lines 131 and 137) omit `nullable=True`. PyMilvus passes the flag through as a kwarg and only raises client-side for vector fields, so these scalar calls fail server-side instead — Milvus rejects a non-nullable field added to a collection that already has rows. That is the very first step of the guide, so a copy-paste user is blocked before reaching any backfill code.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+target field IDs required by the backfill:
+
+```python
+client.create_snapshot(
+    collection_name="my_collection",
+    snapshot_name="bkfill_20260417",
+    compaction_protection_seconds=86400,   # pin files for 24h
+)
+
+info = client.describe_snapshot(snapshot_name="bkfill_20260417")
```
liliu-z: The snapshot snippet does not run against the pinned client, and its comment describes the wrong path shape. `describe_snapshot` takes `collection_name` as a required positional argument, so the call raises TypeError; it returns a `SnapshotInfo` object with no `__getitem__`, so `info["s3_location"]` would raise even if the call succeeded. Separately, the comment claims `s3_location` is already a full `s3a://bucket/...` path, while the repo's own demo and integration test treat it as bucket-relative and prepend the bucket. Pass `collection_name`, use attribute access, and show the value in the shape the rest of the repo expects. `scripts/create_snapshot.py:43` has the same two API errors but is untouched by this PR, so fixing it is optional cleanup.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+`segment_statuses`. Check that `failed_segments` is zero instead of relying on
+HTTP status alone. The `result_path` must be readable through Milvus's object
+storage configuration.
+
+### 5.6 (Optional) Build an index on the new field
+
+Backfill does **not** build indexes. If you need one:
+
+```python
+client.create_index(
```
liliu-z, line 283: The `create_index` call at line 272 is shown without `index_params`, which PyMilvus requires. Users following the guide's post-backfill indexing step hit an immediate client-side error and cannot complete the documented workflow.

src/main/scala/operations/backfill/BackfillResult.scala
```diff
@@ -43,8 +43,9 @@ case class SegmentBackfillResult(
       * kept separate so consumers don't have to rely on that invariant.
       */
     sourceRowCount: Long = 0L,
-    /** Source rows whose PK was matched in the backfill parquet. Derived from a
-      * `__bf_matched__` marker column injected before the left join.
+    /** Source rows whose resolved join key was matched in the backfill parquet.
```
liliu-z, line 46: This PR makes many-to-one joins legal — only the backfill side is checked for uniqueness, and `BackfillJoinKey` documents that source values may repeat — so `matchedRowCount` now counts fanned-out source rows while `dataFileRate` (line 97) still divides it by the data-file row count. With a repeating physical key the summary prints rates above 100% (100 source rows matching one parquet row renders as `10000.00%`). Only the `summary` string is affected; `toJson` emits raw counts. Base the rate on distinct matched keys, or drop the data-file denominator now that it no longer bounds the numerator.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,417 @@
...
+| --------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
+| `InvalidSnapshot: missing manifest_list and storagev2_manifest_list` | Snapshot was produced before storage-V2/V3 landed on that collection, or the snapshot file is truncated. Re-run `CreateSnapshot`. |
+| `Join-key type mismatch`                            | Parquet key type doesn't exactly match the selected snapshot field. Cast it in your ETL.                        |
+| `duplicate join-key values` / `null join key`       | The parquet join key is invalid. Deduplicate it and remove null-key rows before retrying.                          |
+| `Physical join-key field ... was not found`         | `--join-key` is case-sensitive and must name a persisted snapshot field.                                        |
+| `Field name X not found in collection schema`       | Your column-mapping targets a field that isn't in the collection. Verify schema and mapping.                    |
```
liliu-z: The troubleshooting table is indexed by error text so users can match what they see in the log, but this row's key does not correspond to any message the connector emits. The actual error for a column-mapping target missing from the schema is `Fields not found in snapshot schema: ...`. Users grepping for the documented string find nothing.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -133,6 +306,16 @@ object MilvusBackfill {
     // Step 2: Create Milvus client only if no snapshot is available
     var client: MilvusClient = null
     if (snapshotMetadataOpt.isEmpty) {
+      config.joinKey match {
```
yhmo, line 366: src/main/scala/operations/backfill/MilvusBackfill.scala line:309
Medium ---- The physical join-key path has no run()-level regression test: MilvusBackfillTest is unchanged in this PR, so nothing exercises snapshot-based physical-key resolution through the source read (ReaderFieldIDs ordering), the requires-snapshot guard on this block, or processSegments with a physical key. The plan (docs/plans/2026-08-05-backfill-configurable-join-key.md, Task 8 Step 1) explicitly required integration-level tests for this orchestration; the new tests cover resolveJoinKey/prepareBackfillData/performJoin/buildSourceReadProjection in isolation but the wiring is only code-reasoned. Consider an end-to-end regression (physical key + no PK in schema, each mode) or at least a unit test asserting run() rejects a physical key without a snapshot.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -592,25 +740,111 @@ object MilvusBackfill {
...
-    *
-    * PR1 still resolves exactly one primary-key component. The sequence-based
-    * shape is intentional so the read and join stages do not need another
-    * signature change when a later PR adds physical or composite keys.
+  /** Apply column mapping for an explicitly selected physical join field.
+    * Unlike the legacy primary-key path, an omitted mapping never treats a
+    * literal `pk` column specially: the input must contain the exact selected
+    * field name.
+    */
+  private[backfill] def applyPhysicalJoinColumnMapping(
```
yhmo, line 872: src/main/scala/operations/backfill/MilvusBackfill.scala line:748
Low ---- applyPhysicalJoinColumnMapping duplicates most of applyColumnMapping's validation: missing source columns, duplicate targets, join-key/PK-target requirement, at-least-one-target check, and the single-pass aliased select. These two siblings will drift when shared rules change (e.g., new reserved-name handling or message wording). Consider extracting the common mapping validation into one helper and keeping only the legacy literal-`pk` rename in applyColumnMapping.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -226,51 +404,23 @@ object MilvusBackfill {
...
-          .getOrElse(
-            return Left(
-              SchemaValidationError(
-                s"Field '$name' not found in snapshot collection schema"
-              )
-            )
-          )
-        name -> field
-      }.toMap
+      val targetFieldsByName = resolveBackfillTargetFields(
```
yhmo, line 473: src/main/scala/operations/backfill/MilvusBackfill.scala line:414
Medium ---- This is the right place to reject target names that collide (under Spark's configured resolver) with every internal column generated later. The alias allocation already handles __BF_JOIN_0__ variants, but run() only rejects an exact __bf_matched__ target, and coalesce/overwrite additionally create __bf_used_src_<field>__, __bf_used_bf_<field>__, and <field>__bf__ without collision checks. With the default case-insensitive resolver, a valid target such as __BF_MATCHED__ can therefore be overwritten or become ambiguous and reach the writer projection as marker data instead of its declared value. Please reserve all generated names using the configured resolver before source reads and add case-variant/provenance/suffix tests.

docs/user-guide-snapshot-backfill.md
```diff
@@ -0,0 +1,426 @@
...
+
+```text
+--join-key external_row_id \
+--column-mapping source_row_id:external_row_id,category:category,score:score
+```
+
+The physical field must exist exactly (including case) in the snapshot schema
+and must be declared non-nullable. The parquet key must also be non-null and
+unique so one source row cannot fan out into multiple output rows. Source
+values may repeat; the same parquet record is applied to every matching
```
yhmo, line 100: docs/user-guide-snapshot-backfill.md line:100
Low ---- The implementation, tests, README, and this guide now intentionally allow repeated source keys, but the PR description and docs/plans/2026-08-05-backfill-configurable-join-key.md still require non-null uniqueness on both sides and source-side duplicate rejection. A future maintainer could reintroduce source cardinality validation and reject valid many-to-one jobs by trusting that plan. Please update the checked-in plan (and PR description) to record the accepted contract, or restore validation if one-to-one is still intended.

LawrenceTL92 (approved):
Brought the branch up to date with main (two merges: after #116 and after #117; `--input-format` is now derived inside `buildConfig` alongside `--join-key`) and addressed the open review threads:

- **Internal column reservation** (yhmo, MilvusBackfill.scala:414): `validateTargetNamesAgainstInternalColumns` now reserves the match flag, join aliases, per-field provenance flags and the `__bf__` suffix under `spark.sessionState.conf.resolver` before any source read, replacing the exact-case `__bf_matched__` check. Tests cover case variants, provenance names, suffix names, join aliases and a case-sensitive resolver.
- **run()-level guard** (yhmo, MilvusBackfill.scala:309): added `MilvusBackfill.run rejects a physical join key without a snapshot`. A full end-to-end physical-key run needs a live Milvus and belongs in `src/it` (now split out by #116); left as a follow-up.
- **Plan/PR description drift** (yhmo, user guide:100): the checked-in plan now records the accepted contract (backfill-side unique + non-null, source-side may repeat) and the PR description was updated to match.
- **applyPhysicalJoinColumnMapping duplication** (yhmo, :748, Low): not addressed here; it's a maintainability cleanup with no behavioral impact and I'd rather not widen this PR further.

Verified locally: `testOnly com.zilliz.spark.connector.operations.backfill.*` → 181/181 pass; scalafmt clean for all touched files.

PR #110, opened 2026-08-13
fix(backfill): support Alibaba OSS paths

tsunghanjacktsai:
## Summary
- route parquet, snapshot, StorageV2 manifest, and result JSON paths through OSS when `s3CloudProvider=aliyun`
- configure Hadoop OSS without overwriting the platform-injected OSS credentials provider
- preserve managed IAM role propagation and add OSS path/configuration coverage

## Validation
- scalafmt completed
- targeted Cloud-side contract tests pass
- Spark compile/test is environment-blocked locally until the required milvus-storage JNI jar is built; the failure is missing generated `io.milvus.storage` classes, not this change

## Dependency
The matching zilliz-cloud PR passes `--s3-cloud-provider aliyun` and must be deployed with this connector change.

### DIFF
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -463,11 +463,11 @@ object MilvusBackfill {
     // and per-bucket fs.s3a.bucket.<b>.* config is only honored by
     // S3AFileSystem. Force the s3a scheme so the credentials we just wrote
     // actually take effect.
-    val path = normalizeS3Scheme(rawPath)
+    val path = normalizeObjectStorageScheme(rawPath)
     try {
       // Ensure Hadoop S3A is configured for the source bucket (may differ
       // from the Milvus storage bucket) before reading the parquet.
-      configureHadoopS3ForPath(spark, path, config, isSource = true)
+      configureHadoopStorageForPath(spark, path, config, isSource = true)
       val df = spark.read.parquet(path)
 
       // Minimum shape check; pk presence and field-count checks are enforced
@@ -1590,9 +1590,9 @@ object MilvusBackfill {
       // Register S3A bucket credentials so V2SegmentLoader can read AVRO +
       // parquet footers from minio / S3. The main bucket (not the source
       // parquet bucket) holds the snapshot.
-      configureHadoopS3ForPath(
+      configureHadoopStorageForPath(
         spark,
-        s"s3a://${config.s3BucketName}/",
+        storagePath(config, ""),
         config,
         isSource = false
       )
@@ -1603,7 +1603,10 @@ object MilvusBackfill {
           config.s3BucketName,
           hadoopConf,
           manifestSchemaVersion = metadata.manifestSchemaVersion,
-          applyDeletes = false
+          applyDeletes = false,
+          storageScheme =
+            if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun")) "oss"
+            else "s3a"
         ) match {
         case Right(segs) =>
           // Workaround for Milvus snapshot not yet exposing FieldBinlog.child_fields:
@@ -1707,7 +1710,8 @@ object MilvusBackfill {
     val allocator: () => Long = () => logIdCounter.incrementAndGet()
 
     val outputRoot =
-      s"s3a://${config.s3BucketName}/${config.s3RootPath.stripSuffix("/")}/insert_log/" +
+      s"${if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun")) "oss"
+        else "s3a"}://${config.s3BucketName}/${config.s3RootPath.stripSuffix("/")}/insert_log/" +
         s"$collectionID/$partitionID/$segmentID"
 
     // Reuse the same MilvusOption plumbing the V3 writer uses — the new V2
@@ -1927,9 +1931,9 @@ object MilvusBackfill {
     // Same s3:// → s3a:// normalization as readSnapshotJson /
     // readBackfillData. Without it, an s3:// URL would route to the legacy
     // S3FileSystem which ignores fs.s3a.bucket.<b>.* config.
-    val outputPath = normalizeS3Scheme(rawOutputPath)
+    val outputPath = normalizeObjectStorageScheme(rawOutputPath)
     try {
-      configureHadoopS3ForPath(spark, outputPath, config, isSource = false)
+      configureHadoopStorageForPath(spark, outputPath, config, isSource = false)
       val hadoopPath = new Path(outputPath)
       val fs = hadoopPath.getFileSystem(spark.sparkContext.hadoopConfiguration)
       val out = fs.create(hadoopPath, true)
@@ -1969,6 +1973,90 @@ object MilvusBackfill {
     if (path == null) null
     else if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://")
     else path
+  }
+
+  private[backfill] def normalizeObjectStorageScheme(path: String): String =
+    normalizeS3Scheme(path)
+
+  private[backfill] def storagePath(
+      config: BackfillConfig,
+      suffix: String
+  ): String = {
+    val scheme =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun")) "oss://"
+      else "s3a://"
+    scheme + config.s3BucketName + "/" + suffix.stripPrefix("/")
+  }
+
+  private[backfill] def configureHadoopStorageForPath(
+      spark: SparkSession,
+      path: String,
+      config: BackfillConfig,
+      isSource: Boolean
+  ): Unit = {
+    val normalized = normalizeObjectStorageScheme(path)
+    if (normalized != null && normalized.startsWith("oss://")) {
+      configureHadoopOssForPath(spark, normalized, config, isSource)
+    } else {
+      configureHadoopS3ForPath(spark, normalized, config, isSource)
+    }
+  }
+
+  private[backfill] def configureHadoopOssForPath(
+      spark: SparkSession,
+      path: String,
+      config: BackfillConfig,
+      isSource: Boolean
+  ): Unit = {
+    if (path == null || !path.startsWith("oss://")) return
+    val withoutScheme = path.stripPrefix("oss://")
+    val slash = withoutScheme.indexOf('/')
+    val bucket =
+      if (slash < 0) withoutScheme else withoutScheme.substring(0, slash)
+    if (bucket.isEmpty) return
+
+    val endpoint =
+      if (isSource) config.sourceS3Endpoint.getOrElse(config.s3Endpoint)
+      else config.s3Endpoint
+    val accessKey =
+      if (isSource) config.sourceS3AccessKey.getOrElse(config.s3AccessKey)
+      else config.s3AccessKey
+    val secretKey =
+      if (isSource) config.sourceS3SecretKey.getOrElse(config.s3SecretKey)
+      else config.s3SecretKey
+    val useSSL =
+      if (isSource) config.sourceS3UseSSL.getOrElse(config.s3UseSSL)
+      else config.s3UseSSL
+    val useIam =
+      if (isSource) config.sourceS3UseIam.getOrElse(config.s3UseIam)
+      else config.s3UseIam
+    val hadoopConf = spark.sparkContext.hadoopConfiguration
+    hadoopConf.set(
+      "fs.oss.impl",
+      "org.apache.hadoop.fs.aliyun.oss.AliyunOSSFileSystem"
+    )
+    if (endpoint != null && endpoint.nonEmpty)
+      hadoopConf.set("fs.oss.endpoint", endpoint)
+    hadoopConf.set("fs.oss.connection.secure.enabled", useSSL.toString)
+    if (useIam) {
+      val provider = Option(
+        hadoopConf.getTrimmed(BackfillConfig.HadoopOssCredentialsProvider)
+      ).filter(_.nonEmpty)
+      if (provider.isEmpty) {
+        throw new IllegalArgumentException(
+          s"${BackfillConfig.HadoopOssCredentialsProvider} must be configured for OSS IAM access"
+        )
+      }
+      hadoopConf.unset("fs.oss.accessKeyId")
+      hadoopConf.unset("fs.oss.accessKeySecret")
+    } else {
+      hadoopConf.set("fs.oss.accessKeyId", accessKey)
+      hadoopConf.set("fs.oss.accessKeySecret", secretKey)
+      hadoopConf.set(
+        BackfillConfig.HadoopOssCredentialsProvider,
+        "org.apache.hadoop.fs.aliyun.oss.AliyunCredentialsProvider"
+      )
+    }
   }
 
   /** Configure Hadoop S3A credentials for the bucket referenced by `path`.
@@ -2101,12 +2189,14 @@ object MilvusBackfill {
     try {
       // Check if it's an S3 path
       if (
-        snapshotPath.startsWith("s3://") || snapshotPath.startsWith("s3a://")
+        snapshotPath.startsWith("s3://") || snapshotPath.startsWith(
+          "s3a://"
+        ) || snapshotPath.startsWith("oss://")
       ) {
 
         // Construct full S3 path (ensure s3a:// scheme for Hadoop)
         val s3Path = if (snapshotPath.startsWith("s3://")) {
-          normalizeS3Scheme(snapshotPath)
+          normalizeObjectStorageScheme(snapshotPath)
         } else {
           snapshotPath
         }
@@ -2114,7 +2204,7 @@ object MilvusBackfill {
         // Configure S3 settings on Spark's Hadoop Configuration (per-bucket
         // so that snapshot bucket and backfill source bucket can use
         // different credentials in the same Spark session).
-        configureHadoopS3ForPath(spark, s3Path, config, isSource = false)
+        configureHadoopStorageForPath(spark, s3Path, config, isSource = false)
 
         // Use Spark's DataFrame API to read the file (avoids Hadoop version issues)
         val df = spark.read.text(s3Path)
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -4,7 +4,8 @@ import scala.collection.mutable
 import scala.collection.JavaConverters._
 
 import org.apache.hadoop.fs.Path
-import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.connector.write.DataWriter
 import org.apache.spark.sql.functions._
@@ -36,6 +37,11 @@ import com.zilliz.spark.connector.write.{
 object MilvusBackfill {
 
   private val logger = LoggerFactory.getLogger(getClass)
+
+  private[backfill] final case class BackfillSource(
+      dataFrame: DataFrame,
+      checkpointRDD: Option[RDD[Row]]
+  )
 
   /** Marker column added to the backfill side before the left join so we can
     * count how many source rows found a join-key match (non-null marker →
@@ -104,9 +110,11 @@ object MilvusBackfill {
       case Right(_) => // Continue
     }
 
-    // Tracked so we can unpersist in the outer finally, even on mid-flight
-    // failure (the cache is taken right after column mapping, below).
-    var cachedSourceDF: DataFrame = null
+    // OSS source reads use a local-checkpointed RDD to sever the external
+    // storage lineage while scoped credentials are installed. Keep the actual
+    // persisted RDD so its blocks can be released after the transformed
+    // backfill DataFrame is materialized, including on failure.
+    var sourceCheckpointRDD: Option[RDD[Row]] = None
     var cachedBackfillDF: DataFrame = null
 
     logger.info(s"Backfill mode: ${config.mode}")
@@ -191,9 +199,9 @@ object MilvusBackfill {
       val rawBackfillDF =
         readBackfillData(spark, backfillDataPath, config) match {
           case Left(error) => return Left(error)
-          case Right(df) =>
-            cachedSourceDF = df
-            df
+          case Right(source) =>
+            sourceCheckpointRDD = source.checkpointRDD
+            source.dataFrame
         }
 
       // Reproject via the existing PK column-mapping contract, then separate
@@ -304,8 +312,8 @@ object MilvusBackfill {
         case Right(stats) => stats
       }
       val backfillRowCount = joinKeyStats.rowCount
-      cachedSourceDF.unpersist()
-      cachedSourceDF = null
+      sourceCheckpointRDD.foreach(_.unpersist())
+      sourceCheckpointRDD = None
       logger.info(
         s"Backfill data file rows: $backfillRowCount " +
           s"(distinct join keys: ${joinKeyStats.distinctValidKeyCount})"
@@ -455,12 +463,12 @@ object MilvusBackfill {
             logger.warn("Failed to unpersist backfill DataFrame", e)
         }
       }
-      if (cachedSourceDF != null) {
+      sourceCheckpointRDD.foreach { checkpointRDD =>
         try {
-          cachedSourceDF.unpersist()
+          checkpointRDD.unpersist()
         } catch {
           case e: Exception =>
-            logger.warn("Failed to unpersist source backfill DataFrame", e)
+            logger.warn("Failed to unpersist source checkpoint RDD", e)
         }
       }
     }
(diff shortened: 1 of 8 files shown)


src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
+        throw new IllegalArgumentException(
+          s"${BackfillConfig.HadoopOssCredentialsProvider} must be configured for OSS IAM access"
+        )
+      }
+      hadoopConf.unset("fs.oss.accessKeyId")
+      hadoopConf.unset("fs.oss.accessKeySecret")
+    } else {
+      hadoopConf.set("fs.oss.accessKeyId", accessKey)
+      hadoopConf.set("fs.oss.accessKeySecret", secretKey)
+      hadoopConf.set(
```
liliu-z: Setting `fs.oss.credentials.provider` to `org.apache.hadoop.fs.aliyun.oss.AliyunCredentialsProvider` makes OSS filesystem init throw. hadoop-aliyun 3.4.1 instantiates a configured provider class by trying a `(URI, Configuration)` constructor and then a no-arg constructor; `AliyunCredentialsProvider` has neither (only `(Configuration)`), so `AliyunOSSFileSystemStore.initialize` rethrows as `IOException("... constructor exception ...")` before any object is read. This is reachable straight from the documented CLI: `BackfillApp` sets `useIam=false` whenever `--s3-access-key`/`--s3-secret-key` are passed, so every static-credential aliyun run dies at the first `oss://` access. Leaving the key unset instead makes `AliyunOSSUtils.getCredentialsProvider` take the `StringUtils.isEmpty(className)` branch and construct `new AliyunCredentialsProvider(conf)` from the `fs.oss.accessKeyId`/`accessKeySecret` this code already writes.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
     else path
   }
 
+  private[backfill] def normalizeObjectStorageScheme(path: String): String =
```
liliu-z: `normalizeObjectStorageScheme` just delegates to `normalizeS3Scheme`, so it maps `s3://` to `s3a://` regardless of provider and never produces `oss://`. That leaves the two halves of this PR disagreeing: `V2SegmentLoader.resolvePath` now rewrites `s3://` to the provider scheme, but the parquet, snapshot and result-JSON paths that flow through this helper stay on s3a and get dispatched to `configureHadoopS3ForPath`/S3AFileSystem. It matters because the backfill README's own invocation example uses `--parquet s3a://... --snapshot s3a://...`; an aliyun user following it gets AWS S3A wiring pointed at an OSS endpoint and the job fails. Either make this helper provider-aware (it would need the config passed in) or drop the alias and document that only explicit `oss://` inputs are routed to OSS.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
+      if (isSource) config.sourceS3SecretKey.getOrElse(config.s3SecretKey)
+      else config.s3SecretKey
+    val useSSL =
+      if (isSource) config.sourceS3UseSSL.getOrElse(config.s3UseSSL)
+      else config.s3UseSSL
+    val useIam =
+      if (isSource) config.sourceS3UseIam.getOrElse(config.s3UseIam)
+      else config.s3UseIam
+    val hadoopConf = spark.sparkContext.hadoopConfiguration
+    hadoopConf.set(
```
liliu-z, line 2147: The OSS branch writes global `fs.oss.*` keys, so it drops the per-bucket credential isolation the S3A branch provides and the README promises. Concretely: an aliyun run with `--source-s3-access-key`/`--source-s3-secret-key` gives `sourceS3UseIam=Some(false)`, so the `isSource=true` call takes the static branch and overwrites the global `fs.oss.credentials.provider` that the platform injected for the managed role; the later `isSource=false` call then sees a non-empty provider, passes its guard, unsets the access keys, and the main-bucket OSS access ends up with the wrong provider. Note the reviewer's suggested fix does not work — hadoop-aliyun 3.4.1 has no `fs.oss.bucket.<name>.*` mechanism at all — so isolation has to come from a per-access `Configuration` clone (plus `fs.oss.impl.disable.cache=true`, which `V2SegmentLoader.readAllBytes` already honours) or from refusing mixed static/IAM OSS configs outright.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
+      "fs.oss.impl",
+      "org.apache.hadoop.fs.aliyun.oss.AliyunOSSFileSystem"
+    )
+    if (endpoint != null && endpoint.nonEmpty)
+      hadoopConf.set("fs.oss.endpoint", endpoint)
+    hadoopConf.set("fs.oss.connection.secure.enabled", useSSL.toString)
```
liliu-z: `fs.oss.connection.secure.enabled` is written unconditionally from `useSSL`, which defaults to false, so it overrides hadoop-aliyun's own HTTPS default and any run that doesn't pass `--s3-use-ssl` talks to OSS over plaintext HTTP (including the STS security token in the signed headers when a role provider is in play). This mirrors what the S3A branch already does with `fs.s3a.connection.ssl.enabled`, so it is a consistency choice rather than a new regression, but for a public `oss-cn-*.aliyuncs.com` endpoint the safer shape is to only set the key when `useSSL` is true and otherwise let the library default stand.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
+      else config.s3SecretKey
+    val useSSL =
+      if (isSource) config.sourceS3UseSSL.getOrElse(config.s3UseSSL)
+      else config.s3UseSSL
+    val useIam =
+      if (isSource) config.sourceS3UseIam.getOrElse(config.s3UseIam)
+      else config.s3UseIam
+    val hadoopConf = spark.sparkContext.hadoopConfiguration
+    hadoopConf.set(
+      "fs.oss.impl",
```
liliu-z, line 2148: `fs.oss.impl` is set to a class this repo never puts on the classpath. `hadoop-aliyun` appears nowhere in `project/Dependencies.scala`, `build.sbt`, the Dockerfile, or any README/script, so the only reason this works is that the deployment image happens to ship it. Because the class is referenced as a string it compiles fine and fails at runtime with `ClassNotFoundException` / `No FileSystem for scheme: oss`. Add `hadoop-aliyun` (provided/test, alongside `hadoopAws`) or document it as an explicit standalone prerequisite.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
+    )
+    if (endpoint != null && endpoint.nonEmpty)
+      hadoopConf.set("fs.oss.endpoint", endpoint)
+    hadoopConf.set("fs.oss.connection.secure.enabled", useSSL.toString)
+    if (useIam) {
+      val provider = Option(
+        hadoopConf.getTrimmed(BackfillConfig.HadoopOssCredentialsProvider)
+      ).filter(_.nonEmpty)
+      if (provider.isEmpty) {
+        throw new IllegalArgumentException(
```
liliu-z, line 2160: In IAM mode the OSS branch aborts unless the platform has already injected `fs.oss.credentials.provider`, and `BackfillApp` auto-enables IAM whenever AK/SK are omitted — so a self-managed aliyun run that relies on an ECS RAM role or env credentials fails immediately. Fail-fast is a reasonable choice here (hadoop-aliyun ships no default IAM chain to fall back to, and the message names the exact key), but the backfill README currently promises the opposite for this mode, so the docs need updating alongside this.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +1975,90 @@ object MilvusBackfill {
...
 
+  private[backfill] def normalizeObjectStorageScheme(path: String): String =
+    normalizeS3Scheme(path)
+
+  private[backfill] def storagePath(
+      config: BackfillConfig,
+      suffix: String
+  ): String = {
+    val scheme =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun")) "oss://"
```
liliu-z: The `s3CloudProvider == "aliyun" ? oss : s3a` decision is written out three times in this PR. Pull it into one helper (`storageScheme(config)`) and have the other two sites use it — the outputRoot at 1713 can be built from `storagePath` directly.

src/test/scala/operations/backfill/BackfillAppTest.scala
```diff
@@ -494,6 +494,58 @@ class BackfillAppTest extends AnyFunSuite with Matchers with BeforeAndAfterAll {
...
+        isSource = false
+      )
+      hc.get(
+        "fs.oss.impl"
+      ) shouldBe "org.apache.hadoop.fs.aliyun.oss.AliyunOSSFileSystem"
+      hc.get("fs.oss.endpoint") shouldBe "oss-cn-hangzhou-internal.aliyuncs.com"
+      hc.get(BackfillConfig.HadoopOssCredentialsProvider) shouldBe provider
+      hc.get("fs.oss.accessKeyId") shouldBe null
+      hc.get("fs.oss.accessKeySecret") shouldBe null
+    } finally {
```
liliu-z, line 564: The `finally` here only restores `fs.oss.credentials.provider`; the three keys `configureHadoopOssForPath` writes — `fs.oss.impl`, `fs.oss.endpoint`, `fs.oss.connection.secure.enabled` — stay on the shared session configuration for every test that runs afterwards. These are global (not bucket-scoped like the pre-existing `fs.s3a.bucket.*` leakage), and `fs.oss.impl` names a class that isn't on the test classpath, so a future test touching an `oss://` path would fail with a confusing ClassNotFoundException. Unset all four.

src/main/scala/operations/backfill/MilvusBackfill.scala
liliu-z: Only the IAM branch of `configureHadoopOssForPath` has a test; the static AK/SK branch — the one that writes the unconstructible provider class — has none. A test asserting the Hadoop conf this branch produces (and ideally attempting an FS init) would have caught it.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -52,12 +52,13 @@ object V2SegmentLoader extends Logging {
       bucket: String,
       hadoopConf: Configuration,
       manifestSchemaVersion: Int = 1,
-      applyDeletes: Boolean = true
+      applyDeletes: Boolean = true,
+      storageScheme: String = "s3a"
```
liliu-z, line 58: `storageScheme` defaults to `"s3a"`, so the two other callers of `loadV2Segments` compile unchanged and stay on s3a. That means an Alibaba-hosted StorageV2 collection read through the connector's own DataSource path still resolves bucket-relative manifest and binlog paths to `s3a://`, even though the backfill path now resolves the same paths to `oss://`. If keeping the read path on s3a is intentional PR scoping, say so in the scaladoc; otherwise these call sites need the provider threaded through too.

tsunghanjacktsai (commented):
Implemented the requested OSS blocker fixes in 1e781c2. Detailed replies will be added to each inline thread.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. Static OSS mode now writes only fs.oss.accessKeyId/fs.oss.accessKeySecret and leaves fs.oss.credentials.provider unset, so hadoop-aliyun constructs AliyunCredentialsProvider(Configuration) through its supported path. Added a test that resolves the provider and verifies the AK/SK.

tsunghanjacktsai: Fixed in 1e781c2. Static OSS mode now writes only fs.oss.accessKeyId/fs.oss.accessKeySecret and leaves fs.oss.credentials.provider unset, so hadoop-aliyun constructs AliyunCredentialsProvider(Configuration) through its supported path. Added a test that resolves the provider and verifies the AK/SK.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. normalizeObjectStorageScheme is now provider-aware: Alibaba maps both s3:// and s3a:// inputs to oss://, while AWS keeps the existing s3:// to s3a:// behavior. Parquet, snapshot, and result JSON paths all use this helper, with tests for both Alibaba aliases.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. OSS access is now scoped per operation: all fs.oss.* values are saved, source or target settings are installed with filesystem caching disabled, and the managed configuration is restored in finally. Source parquet is eagerly local-checkpointed inside the source credential scope so its external OSS lineage cannot be evaluated later under target credentials. StorageV2 manifest reads use a cloned Hadoop Configuration.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. OSS now sets fs.oss.connection.secure.enabled only when SSL is explicitly enabled. The false/default case unsets the override and preserves hadoop-aliyun HTTPS defaults.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. Added hadoop-aliyun 3.4.1 as provided,test alongside hadoop-aws. README also states that the Spark runtime image must provide this module, since it is intentionally not bundled into the assembly to avoid Hadoop runtime conflicts.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. README no longer claims a generic default IAM chain for Alibaba OSS. It now explicitly requires the managed runtime to inject fs.oss.credentials.provider and documents that hadoop-aliyun has no generic ECS/env fallback for this access path.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. The provider-to-scheme decision is centralized in storageScheme(config), and storagePath(config, suffix) now builds all target paths from that helper. The duplicated Alibaba conditionals were removed.

src/test/scala/operations/backfill/BackfillAppTest.scala
tsunghanjacktsai: Fixed in 1e781c2. Test cleanup now unsets every OSS key touched by configuration, including fs.oss.impl, fs.oss.impl.disable.cache, fs.oss.endpoint, fs.oss.connection.secure.enabled, fs.oss.accessKeyId, fs.oss.accessKeySecret, and fs.oss.credentials.provider.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 1e781c2. Added static AK/SK coverage that verifies the generated Hadoop configuration and calls AliyunOSSUtils.getCredentialsProvider to prove filesystem credential initialization succeeds with the configured keys.

src/main/scala/read/V2SegmentLoader.scala
tsunghanjacktsai: This is intentional for PR #110 scope. The generic connector DataSource and existing tools do not expose a cloud-provider option, so their default storageScheme remains s3a for backward compatibility. Alibaba Backfill explicitly passes storageScheme="oss" and supplies the OSS Hadoop configuration. The scaladoc now documents this contract so the default is not mistaken for Alibaba support.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -203,13 +208,23 @@ object V2SegmentLoader extends Logging {
...
-    else if (path.startsWith("s3a://")) path
-    else if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://")
-    else if (bucket != null && bucket.nonEmpty)
-      s"s3a://$bucket/${path.stripPrefix("/")}"
+    else if (path.startsWith("s3a://") || path.startsWith("oss://")) path
+    else if (path.startsWith("s3://")) {
+      val scheme = storageScheme.stripSuffix("://")
+      s"$scheme://" + path.stripPrefix("s3://")
+    } else if (bucket != null && bucket.nonEmpty)
+      s"${storageScheme.stripSuffix("://")}://$bucket/${path.stripPrefix("/")}"
```
congqixia: V2SegmentLoader.resolvePath preserves s3a:// unconditionally. When storageScheme = "oss", fully qualified manifest or binlog paths therefore still use S3A,
     even though the supplied Hadoop configuration is configured for fs.oss.*. Both s3:// and s3a:// should be rewritten to the requested storage scheme. Please
     add coverage for s3a:// manifest and nested binlog paths.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -463,25 +477,26 @@ object MilvusBackfill {
...
-        )
-      }
-
-      Right(df)
+        }
+        // A normal cache keeps the source lineage and may re-read OSS after
+        // eviction, when the scoped source credentials have already been
+        // restored. Eager local checkpointing materializes the input and
+        // truncates that external-storage lineage before leaving this scope.
+        df.localCheckpoint(eager = true)
```
congqixia: df.localCheckpoint(eager = true) persists the underlying RDD with MEMORY_AND_DISK, while cachedSourceDF.unpersist() only operates through Spark SQL’s
     CacheManager and does not unpersist that checkpoint RDD. The transformed backfillDF is then cached separately, potentially retaining two full copies of a
     large backfill dataset for most of the job. This also affects AWS and local-file jobs, although scoped credential materialization is only required for OSS.
     Please restrict this behavior to the OSS path and retain an actual RDD cleanup handle.

src/main/scala/read/V2SegmentLoader.scala
tsunghanjacktsai: Fixed in 6115c9d. resolvePath now treats both s3:// and s3a:// as aliases for the requested storageScheme, while preserving other explicit schemes. Added OSS coverage for an s3a:// manifest and a nested s3a:// insert-log parquet path.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 6115c9d. Local checkpointing is now used only for oss:// sources. The code checkpoints a copied Row RDD, retains that exact persisted RDD as the cleanup handle, and unpersists it immediately after the transformed backfill DataFrame is materialized or from finally on failure. AWS and local-file reads return without a source checkpoint. Added a test that verifies the checkpoint RDD is registered and actually removed.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -170,7 +175,8 @@ object V2SegmentLoader extends Logging {
                   "emit an empty V2ColumnGroup from a partial manifest"
               )
             }
-            val samplePath = resolvePath(afb.binlogs.head.logPath, bucket)
+            val samplePath =
```
congqixia: This normalized path is only used to read the sample parquet footer. The subsequent call to `toV2SegmentInfo(entry, ...)` still copies the original manifest
  paths through `sorted.map(_.logPath)` and `log.logPath`. Therefore, a manifest containing fully qualified `s3a://` binlog paths can pass footer inspection
  through OSS, while the `V2SegmentInfo` later passed to the native packed reader still contains `s3a://`.

  Please normalize every returned `columnGroups.filePaths` and `deltaLogs.logPath`, not only `samplePath`. The test should assert the paths returned by
  `loadV2Segments`, rather than testing `resolvePath` in isolation.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -300,6 +312,8 @@ object MilvusBackfill {
         case Right(stats) => stats
       }
       val backfillRowCount = joinKeyStats.rowCount
+      sourceCheckpointRDD.foreach(_.unpersist())
+      sourceCheckpointRDD = None
```
congqixia: Although the aggregation has materialized `backfillDF.cache()`, that cached DataFrame still has lineage to this local-checkpoint RDD. Local checkpointing has
  already truncated the original OSS lineage. Once these checkpoint blocks are explicitly unpersisted, losing any cached backfill block—for example because of
  executor loss, decommissioning, or dynamic allocation—makes that partition impossible to recompute and can fail the remaining join/write stages with a missing
  checkpoint block.

  Please retain this checkpoint until all consumers have completed, or materialize the final backfill DataFrame into an independent recoverable/checkpointed
  lineage before releasing the source checkpoint.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1971,6 +2028,162 @@ object MilvusBackfill {
...
+
+  private val OssHadoopKeys = Seq(
+    "fs.oss.impl",
+    "fs.oss.impl.disable.cache",
+    "fs.oss.endpoint",
+    "fs.oss.connection.secure.enabled",
+    "fs.oss.accessKeyId",
+    "fs.oss.accessKeySecret",
+    BackfillConfig.HadoopOssCredentialsProvider
+  )
```
congqixia, line 2065: `fs.oss.securityToken` is missing from this saved/restored key set. Hadoop 3.4.1's `AliyunCredentialsProvider` automatically combines a non-empty security
  token with `fs.oss.accessKeyId` and `fs.oss.accessKeySecret`.

  If the managed target configuration contains a session token and the source scope installs separate static AK/SK credentials, the source read will use the
  source AK/SK together with the target token, producing invalid session credentials.

  Please include `fs.oss.securityToken` in `OssHadoopKeys` and unset it when installing long-lived static AK/SK credentials, restoring its previous value when
  the scope exits.

src/main/scala/read/V2SegmentLoader.scala
tsunghanjacktsai: Fixed in 0bc046f. V2SegmentLoader now resolves every nested binlog and delta-log path on the AvroManifestEntry before footer inspection and before toV2SegmentInfo builds the runtime object. The new loadV2Segments test encodes a real manifest, reads a real parquet footer, and asserts the returned columnGroups.filePaths and deltaLogs.logPath use the requested scheme.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 0bc046f. The OSS source checkpoint is now retained until run() finally, so downstream join/write stages keep their lineage root for their full lifetime. To avoid holding two persisted copies, OSS no longer caches the derived backfillDF; only non-OSS sources keep the existing transformed-DataFrame cache. The retained checkpoint RDD is still explicitly unpersisted on every exit path.

src/main/scala/operations/backfill/MilvusBackfill.scala
tsunghanjacktsai: Fixed in 0bc046f. fs.oss.securityToken is now part of the scoped OSS save/restore set. Installing long-lived static source AK/SK explicitly unsets any inherited target token, and the previous managed token is restored when the scope exits. Tests cover IAM preservation, static-token removal, and scoped restoration.

congqixia (approved):
/lgtm

liliu-z (commented):
/lgtm
/approve

xiaofan-luan:
I found two correctness issues that need follow-up:

1. **[P1] Keep Hadoop-resolved paths out of the native packed reader**

   `resolveEntryPaths` converts every manifest binlog/delta path into an `oss://bucket/...` or `s3a://bucket/...` URI, and those strings flow unchanged through `V2SegmentInfo` into `MilvusStorageReader`.

   The pinned milvus-storage revision (`bf7a543a`) treats any absolute URI as an external filesystem path and requires a matching `extfs.*` configuration in [`FilesystemCache::resolve_config`](https://github.com/zilliztech/milvus-storage/blob/bf7a543a2353109912db136ef8bdfd03d8fe4e61/cpp/src/filesystem/fs.cpp#L244-L284). However, `Properties.fromMilvusOption` only supplies `fs.*` properties. Therefore, a normal bucket-relative manifest path is now converted into an absolute URI that the native reader rejects with `No matching external filesystem config`. This affects the new OSS path and also regresses the default S3A Storage V2 path.

   Please resolve paths only for Hadoop reads (manifest/footer access), while keeping the column-group paths passed to milvus-storage bucket-relative, or provide the native URI format together with matching `extfs.*` configuration. Please also add a contract test that reaches `MilvusStorageReader.create`; the current test only verifies the rewritten strings.

2. **[P2] `localCheckpoint()` does not make the OSS source fault-tolerant**

   Retaining the checkpoint RDD gives cleanup code a stable `unpersist()` handle, but it does not make the checkpoint durable. Spark stores local-checkpoint partitions in ephemeral executor-local storage and explicitly documents that executor loss can cause an irrecoverable job failure and that this is unsafe with dynamic allocation: https://github.com/apache/spark/blob/v4.0.0/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L1681-L1695.

   The managed Spark batch template enables dynamic allocation by default. Please use a reliable checkpoint on storage accessible with the target-side credentials, or preserve a recomputable source lineage with an isolated Hadoop credential configuration. Keeping the RDD alive only fixes explicit cleanup; it does not protect against executor loss.

PR #112, opened 2026-08-20
fix: [reader] canonicalize Milvus-format S3 URIs

congqixia:
Milvus servers configured with a custom storage endpoint (non-empty fs.address, e.g. MinIO) write snapshot locations and the data paths embedded in snapshot metadata in the "Milvus format" s3://<address>/<bucket>/<key>. Consumers here assumed bucket-relative (file/...) or standard s3://bucket/key shapes, so the endpoint host was mistaken for the bucket, breaking S3A reads with "object URI bucket ..." mismatches and FileNotFoundException on the snapshot JSON.

Add MilvusSnapshotReader.toStandardS3Path to collapse all three shapes to s3a://bucket/key, and route every snapshot path parse point through it (resolveClientSnapshotLocation, snapshotBucket, buildSnapshotHadoopConf, configureHadoopS3ForPath, readSnapshotJson, V2SegmentLoader.resolvePath, manifest base paths). Detection uses the authority containing a port, since S3 bucket names never contain ':'.

Milvus on real S3 (empty fs.address) is unaffected: URIs there already use the standard shape.

### DIFF
--- a/src/main/scala/MilvusOption.scala
+++ b/src/main/scala/MilvusOption.scala
@@ -475,8 +475,17 @@ case class MilvusS3Option(
 
   def getFilePath(path: String): Path = {
     if (notEmpty(s3FileSystemType)) {
-      if (path.startsWith("s3a://")) {
-        new Path(path)
+      if (path.startsWith("s3a://") || path.startsWith("s3://")) {
+        // Canonicalize Milvus-format s3://<address>/<bucket>/<key> so the
+        // endpoint is not treated as the bucket. Thread the configured
+        // endpoint so port-less endpoints are recognized too.
+        new Path(
+          MilvusStoragePath.toStandardS3Path(
+            path,
+            endpoint = s3Endpoint,
+            configuredBucket = s3BucketName
+          )
+        )
       } else {
         val finalPath = s"s3a://${s3BucketName}/${s3RootPath}/${path}"
         new Path(new URI(finalPath))
--- a/src/main/scala/MilvusStoragePath.scala
+++ b/src/main/scala/MilvusStoragePath.scala
@@ -1,0 +1,410 @@
+package com.zilliz.spark.connector
+
+import java.net.URI
+
+/** Neutral storage-path utilities shared by the config layer (MilvusOption)
+  * and the read layer (MilvusSnapshotReader).
+  *
+  * Milvus storage can surface the same object in three shapes depending on how
+  * the server was configured:
+  *
+  *   - bucket-relative key: `file/snapshots/...` (no scheme)
+  *   - standard S3: `s3://<bucket>/<key>`
+  *   - Milvus-format: `s3://<address>/<bucket>/<key>` (address = storage
+  *     endpoint, embedded whenever `fs.address` is non-empty, e.g. MinIO
+  *     deployments)
+  *
+  * [[toStandardS3Path]] collapses all three to the Hadoop `s3a://<bucket>/<key>`
+  * form; [[toBucketRelativeKey]] reduces a path to the bare object key that
+  * milvus-storage's native FFI resolves against the `fs.*` properties.
+  */
+object MilvusStoragePath {
+
+  /** Probe the scheme of a storage path without throwing on input
+    * `java.net.URI` rejects (a space, an unescaped `%`, brackets).
+    *
+    * `s3://` / `s3a://` prefixes are matched textually so the unparseable
+    * fallbacks in [[toStandardS3Path]] / [[toBucketRelativeKey]] / [[bucketOf]]
+    * stay reachable from callers that first branch on the scheme.
+    */
+  private[connector] def schemeOf(path: String): Option[String] = {
+    if (path == null) None
+    else {
+      val trimmed = path.trim
+      if (trimmed.isEmpty) None
+      else if (trimmed.startsWith("s3://")) Some("s3")
+      else if (trimmed.startsWith("s3a://")) Some("s3a")
+      else {
+        try Option(new URI(trimmed).getScheme).map(_.toLowerCase)
+        catch { case _: Exception => None }
+      }
+    }
+  }
+
+  /** Textually split a URI authority into its (host, port).
+    *
+    * `java.net.URI.getHost` / `getPort` return null / -1 for hosts that
+    * violate RFC 2396 `domainlabel` — e.g. a docker-compose service name with
+    * an underscore like `milvus_minio` — so fall back to parsing the raw
+    * authority: drop through the last `@` (userinfo) and split on the last
+    * `:` (a `[...]:port` IPv6 form keeps its brackets).
+    */
+  private def authorityHostPort(
+      authority: String
+  ): (String, Option[Int]) = {
+    if (authority == null || authority.isEmpty) ("", None)
+    else {
+      val noUser = authority.split("@").last
+      val colon = noUser.lastIndexOf(':')
+      if (colon > 0) {
+        val host = noUser.substring(0, colon)
+        val port = noUser.substring(colon + 1).toIntOption
+        (host, port)
+      } else (noUser, None)
+    }
+  }
+
+  /** Normalize a configured storage endpoint to a bare lowercase host, so it
+    * can be compared against a URI authority host.
+    *
+    * Handles `http://host:port`, `host:port`, and bare-host forms. Unknown
+    * shapes fall back to the trimmed, lowercased input.
+    */
+  private[connector] def storageEndpointHost(endpoint: String): String = {
+    if (endpoint == null) return ""
+    val trimmed = endpoint.trim
+    if (trimmed.isEmpty) return ""
+    val uri =
+      try new URI(if (trimmed.contains("://")) trimmed else "//" + trimmed)
+      catch { case _: Exception => return trimmed.toLowerCase }
+    // URI.getHost preserves case and is null for hosts that violate RFC 2396
+    // (e.g. an underscore), so normalize via the textual authority too.
+    Option(uri.getHost)
+      .map(_.toLowerCase)
+      .orElse(
+        Option(uri.getRawAuthority)
+          .map(a => authorityHostPort(a)._1.toLowerCase)
+          .filter(_.nonEmpty)
+      )
+      .getOrElse(trimmed.toLowerCase)
+  }
+
+  /** True when `uri` is a Milvus-format storage URI whose authority is the
+    * storage endpoint rather than the bucket.
+    *
+    * Signals, in order:
+    *
+    *   - the authority carries a port (`host:port`): an S3 bucket name may only
+    *     contain lowercase letters, digits, `-` and `.`, so it can never carry
+    *     a port. The port is read from `getPort` when the host parses as a
+    *     server-based authority, else derived textually from the raw authority
+    *     (hosts with an underscore fail RFC 2396 and yield getPort == -1).
+    *   - the authority host equals the configured storage endpoint host. This
+    *     covers port-less endpoints (e.g. `s3.amazonaws.com`).
+    *   - an endpoint-spelling-independent signal using the configured bucket:
+    *     when the port and endpoint-host checks cannot decide, an authority
+    *     that equals the configured bucket is standard S3 (the bucket is in
+    *     the authority), while a first path segment equal to the configured
+    *     bucket is Milvus-format (the bucket is embedded after the endpoint).
+    *     This survives spelling differences between what Milvus embedded in
+    *     the URI and what the connector was configured with (regional vs
+    *     global host, IP vs DNS name, an alias behind a proxy).
+    */
+  private def isEndpointPrefixed(
+      uri: URI,
+      endpoint: String,
+      configuredBucket: String
+  ): Boolean = {
+    val rawAuthority = Option(uri.getRawAuthority).getOrElse("")
+    val (textHost, textPort) = authorityHostPort(rawAuthority)
+    val port = if (uri.getPort > 0) uri.getPort else textPort.getOrElse(-1)
+    if (port > 0) {
+      true
+    } else {
+      val host = Option(uri.getHost)
+        .map(_.toLowerCase)
+        .orElse(Some(textHost.toLowerCase).filter(_.nonEmpty))
+        .getOrElse("")
+      if (host.nonEmpty && storageEndpointHost(endpoint) == host) {
+        true
+      } else {
+        val configured = configuredBucket.trim.toLowerCase
+        if (configured.nonEmpty && host.nonEmpty) {
+          val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+          val slash = rawPath.indexOf('/')
+          val firstSegment =
+            (if (slash < 0) rawPath else rawPath.substring(0, slash)).toLowerCase
+          if (host == configured) false // authority IS the bucket → standard
+          else if (host.contains("."))
+            // endpoint-like authority (a DNS name or IP): a first path segment
+            // equal to the configured bucket implies Milvus-format. A
+            // single-label authority such as "archive" or "data-lake" is a
+            // standard bucket, not an endpoint — reading it as Milvus-format
+            // would silently rewrite the path to the configured bucket.
+            firstSegment == configured
+          else false
+        } else false
+      }
+    }
+  }
+
+  /** Textually split an unparseable `s3://` / `s3a://` URI (one that
+    * `java.net.URI` rejects — a space, a brace, a stray `%`) into its
+    * authority, path, host and port.
+    *
+    * `java.net.URI` cannot parse these inputs, but the authority/path split on
+    * `/` still works textually, so Milvus-format detection (a `host:port`
+    * authority, or a first path segment equal to the configured bucket) keeps
+    * working instead of leaving the endpoint as the bucket.
+    */
+  private def unparseableAuthorityParts(
+      trimmed: String
+  ): (String, String, String, Option[Int]) = {
+    val schemeEnd = trimmed.indexOf("://")
+    val rest = if (schemeEnd >= 0) trimmed.substring(schemeEnd + 3) else trimmed
+    val slash = rest.indexOf('/')
+    val authority = if (slash < 0) rest else rest.substring(0, slash)
+    val path = if (slash < 0) "" else rest.substring(slash + 1)
+    val (host, port) = authorityHostPort(authority)
+    (authority, path, host, port)
+  }
+
+  /** Reduce a bare object key from an unparseable `s3://` / `s3a://` URI. */
+  private def unparseableBucketRelativeKey(trimmed: String): String = {
+    val (_, path, _, port) = unparseableAuthorityParts(trimmed)
+    if (port.nonEmpty) {
+      // Milvus-format: scheme://address/bucket/key — strip the bucket segment.
+      // With no path segment after the bucket there is no object key.
+      val slash = path.indexOf('/')
+      if (slash < 0) "" else path.substring(slash + 1)
+    } else {
+      // Standard S3: scheme://bucket/key — everything after the authority.
+      path
+    }
+  }
+
+  /** Extract the bucket from an unparseable `s3://` / `s3a://` URI, so a bare
+    * key stripped from the same input is never left to resolve silently
+    * against the connector's configured bucket.
+    */
+  private def unparseableBucket(trimmed: String): String = {
+    val (authority, path, host, port) = unparseableAuthorityParts(trimmed)
+    if (port.nonEmpty) {
+      // Milvus-format: the first path segment is the bucket.
+      val slash = path.indexOf('/')
+      if (slash < 0) path else path.substring(0, slash)
+    } else if (host.nonEmpty) {
+      // Standard S3: the authority host is the bucket.
+      host
+    } else {
+      authority
+    }
+  }
+
+  /** Canonicalize an unparseable `s3://` / `s3a://` URI to
+    * `s3a://bucket/key`, applying the same Milvus-format detection as the
+    * parseable path.
+    */
+  private def unparseableStandardS3Path(trimmed: String): String = {
+    val (authority, path, _, port) = unparseableAuthorityParts(trimmed)
+    if (port.nonEmpty) {
+      val slash = path.indexOf('/')
+      if (slash < 0) "s3a://" + path
+      else {
+        val bucket = path.substring(0, slash)
+        val key = path.substring(slash + 1)
+        s"s3a://$bucket/$key"
+      }
+    } else s"s3a://$authority/$path"
+  }
+
+  /** Extract the bare bucket-relative object key from a storage path.
+    *
+    * This is the form milvus-storage's native FFI expects: it resolves
+    * scheme-less keys against the `fs.*` properties and rejects any
+    * scheme-bearing path for lack of an `extfs.*` config block. Hadoop APIs
+    * should keep the [[toStandardS3Path]] form instead.
+    *
+    *   - `files/snapshots/...` → `files/snapshots/...`
+    *   - `s3://bucket/files/...` → `files/...`
+    *   - `s3://minio:9000/bucket/files/...` (Milvus-format) → `files/...`
+    *
+    * @param path
+    *   storage path or URI to reduce
+    * @param endpoint
+    *   configured storage endpoint (e.g. `fs.address` / `s3Endpoint`), used to
+    *   recognize endpoint-prefixed Milvus-format URIs
+    */
+  private[connector] def toBucketRelativeKey(
+      path: String,
+      endpoint: String = "",
+      configuredBucket: String = ""
+  ): String = {
+    if (path == null) return null
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch {
+        case _: Exception =>
+          // Unparseable s3 URI: textually split the authority so Milvus-format
+          // detection still applies and a bare key is returned for the native
+          // reader, instead of a scheme-bearing path it would reject.
+          if (trimmed.startsWith("s3://") || trimmed.startsWith("s3a://"))
+            return unparseableBucketRelativeKey(trimmed)
+          else return trimmed
+      }
+    Option(uri.getScheme).map(_.toLowerCase) match {
+      case Some("s3") | Some("s3a") =>
+        // getRawPath preserves percent-encoding; getPath would decode %20 to a
+        // space / %23 to a # and produce a broken object key or URI fragment.
+        // Re-append a literal '?'/'#' so the key is not truncated.
+        val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+        val suffix =
+          Option(uri.getRawQuery).map("?" + _).getOrElse("") +
+            Option(uri.getRawFragment).map("#" + _).getOrElse("")
+        if (isEndpointPrefixed(uri, endpoint, configuredBucket)) {
+          // Milvus-format: scheme://address/bucket/key — the bucket segment
+          // precedes the object key. With no path segment after the bucket
+          // there is no object key.
+          val slash = rawPath.indexOf('/')
+          val key = if (slash < 0) "" else rawPath.substring(slash + 1)
+          key + suffix
+        } else {
+          // Standard S3: scheme://bucket/key — everything after the authority
+          // is the object key.
+          rawPath + suffix
+        }
+      case _ => trimmed
+    }
+  }
+
+  /** Extract the bucket a qualified storage URI names, or `None` when `path`
+    * carries no scheme (bucket-relative key).
+    *
+    * Unlike `MilvusScan.snapshotBucket` this never throws on a non-s3 scheme —
+    * it is used to pin the correct bucket on native partition options before
+    * a path is reduced to a bare key.
+    */
+  private[connector] def bucketOf(
+      path: String,
+      endpoint: String = "",
+      configuredBucket: String = ""
+  ): Option[String] = {
+    if (path == null) return None
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return None
+    val uri =
+      try new URI(trimmed)
+      catch {
+        case _: Exception =>
+          // Unparseable s3 URI: textually split the authority so the bucket is
+          // still extracted — toBucketRelativeKey strips the bucket from the
+          // same input, so giving up here would leave the bare key to resolve
+          // silently against the connector's configured bucket.
+          if (trimmed.startsWith("s3://") || trimmed.startsWith("s3a://"))
+            return Some(unparseableBucket(trimmed))
+          else return None
+      }
+    Option(uri.getScheme).map(_.toLowerCase) match {
+      case Some("s3") | Some("s3a") =>
+        if (isEndpointPrefixed(uri, endpoint, configuredBucket)) {
+          // Milvus-format: scheme://address/bucket/key — the first path
+          // segment is the bucket.
+          val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+          val slash = rawPath.indexOf('/')
+          val bucket = if (slash < 0) rawPath else rawPath.substring(0, slash)
+          Option(bucket).filter(_.nonEmpty)
+        } else {
+          // Standard S3: scheme://bucket/key — the authority host is the bucket.
+          // getHost is null for hosts with e.g. an underscore, so fall back to
+          // the authority and strip userinfo (everything after the last '@')
+          // and any port.
+          Option(uri.getHost).orElse(
+            Option(uri.getRawAuthority)
+              .map(_.split("@").last)
+              .map(_.split(":").head)
+              .filter(_.nonEmpty)
+          )
+        }
+      case _ => None
+    }
+  }
+
+  /** Canonicalize a Milvus storage path to Hadoop S3A path-style
+    * `s3a://<bucket>/<key>`.
+    *
+    * This collapses the three shapes (bucket-relative, standard S3,
+    * Milvus-format) to `s3a://<bucket>/<key>` so downstream Hadoop parsers
+    * (bucket extraction, per-bucket S3A conf, S3A reads) see one shape.
+    * Native milvus-storage consumers should use [[toBucketRelativeKey]]
+    * instead.
+    *
+    * Detection of the Milvus-format uses both the authority carrying a port
+    * and a match against the configured `endpoint` host — see
+    * [[isEndpointPrefixed]].
+    *
+    * @param path
+    *   storage path or URI to canonicalize
+    * @param fallbackBucket
+    *   bucket to prefix when `path` is bucket-relative (no scheme)
+    * @param endpoint
+    *   configured storage endpoint (e.g. `fs.address` / `s3Endpoint`); used to
+    *   recognize endpoint-prefixed Milvus-format URIs
+    */
+  private[connector] def toStandardS3Path(
+      path: String,
+      fallbackBucket: String = "",
+      endpoint: String = "",
+      configuredBucket: String = ""
+  ): String = {
+    if (path == null) return null
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch {
+        case _: Exception =>
+          // Unparseable s3 URI (a space, an unescaped %, [ ]): textually split
+          // the authority so an endpoint-prefixed path still collapses to
+          // s3a://bucket/key instead of leaving the endpoint as the bucket.
+          if (trimmed.startsWith("s3://") || trimmed.startsWith("s3a://"))
+            return unparseableStandardS3Path(trimmed)
+          else return trimmed
+      }
+    Option(uri.getScheme).map(_.toLowerCase) match {
+      case Some("s3") | Some("s3a") =>
+        // getRawAuthority / getRawPath preserve percent-encoding so the
+        // reconstructed URI stays parseable (getPath would turn %20 into a
+        // literal space and %23 into a # fragment). getRawPath stops at a
+        // literal '?' / '#', so re-append the raw query / fragment — a global
+        // prefix rewrite would otherwise leave the endpoint as the bucket for
+        // Milvus-format URIs.
+        val authority = Option(uri.getRawAuthority).map(_.trim).filter(_.nonEmpty)
+        val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+        val suffix =
+          Option(uri.getRawQuery).map("?" + _).getOrElse("") +
+            Option(uri.getRawFragment).map("#" + _).getOrElse("")
+        if (isEndpointPrefixed(uri, endpoint, configuredBucket)) {
+          // Milvus-format: scheme://address/bucket/key — the first path
+          // segment is the bucket, the authority is the storage endpoint.
+          val slash = rawPath.indexOf('/')
+          val reconstructed =
+            if (slash < 0) {
+              s"s3a://$rawPath"
+            } else {
+              val bucket = rawPath.substring(0, slash)
+              val key = rawPath.substring(slash + 1)
+              s"s3a://$bucket/$key"
+            }
+          reconstructed + suffix
+        } else {
+          // Standard S3: scheme://bucket/key (also normalize s3 -> s3a).
+          s"s3a://${authority.getOrElse("")}/$rawPath" + suffix
+        }
+      case None if fallbackBucket.trim.nonEmpty =>
+        s"s3a://${fallbackBucket.trim}/${trimmed.stripPrefix("/")}"
+      case _ => trimmed
+    }
+  }
+}
--- a/src/main/scala/operations/backfill/MilvusBackfill.scala
+++ b/src/main/scala/operations/backfill/MilvusBackfill.scala
@@ -15,7 +15,8 @@ import org.slf4j.LoggerFactory
 import com.zilliz.spark.connector.{
   MilvusClient,
   MilvusConnectionParams,
-  MilvusOption
+  MilvusOption,
+  MilvusStoragePath
 }
 import com.zilliz.spark.connector.read.{
   MilvusSnapshotReader,
@@ -393,7 +394,12 @@ object MilvusBackfill {
       val (collectionID, segmentToPartitionMap, segmentBasePathMap) =
         snapshotMetadataOpt match {
           case Some(metadata) =>
-            extractMetadataFromSnapshot(metadata, v2Segments)
+            extractMetadataFromSnapshot(
+              metadata,
+              v2Segments,
+              config.s3Endpoint,
+              config.s3BucketName
+            )
           case None =>
             val (colId, segPartMap) =
               retrieveMilvusMetadata(config, client) match {
@@ -484,9 +490,22 @@ object MilvusBackfill {
   ): Either[BackfillError, BackfillSource] = {
     // Hadoop 3.4.1 has separate s3:// and s3a:// FileSystem implementations
     // and per-bucket fs.s3a.bucket.<b>.* config is only honored by
-    // S3AFileSystem. Force the s3a scheme so the credentials we just wrote
-    // actually take effect.
-    val path = normalizeObjectStorageScheme(rawPath, config)
+    // S3AFileSystem. Force the s3a scheme (and canonicalize any Milvus-format
+    // endpoint-prefixed path, using the source bucket's endpoint override when
+    // present) so the credentials we just wrote actually take effect; Aliyun
+    // OSS paths keep their own scheme.
+    val path =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun"))
+        normalizeObjectStorageScheme(rawPath, config)
+      else
+        // The source bucket is not configurable, so pass "" as the detection
+        // hint: using the Milvus target bucket here would misclassify a source
+        // key whose first segment happens to equal the target bucket name.
+        normalizeS3Scheme(
+          rawPath,
+          config.sourceS3Endpoint.getOrElse(config.s3Endpoint),
+          ""
+        )
     try {
       // Configure the source bucket before reading the parquet. OSS must be
       // fully materialized while this temporary credential scope is active;
@@ -1658,7 +1677,8 @@ object MilvusBackfill {
           hadoopConf,
           manifestSchemaVersion = metadata.manifestSchemaVersion,
           applyDeletes = false,
-          storageScheme = storageScheme(config)
+          storageScheme = storageScheme(config),
+          endpoint = config.s3Endpoint
         ) match {
         case Right(segs) =>
           // Workaround for Milvus snapshot not yet exposing FieldBinlog.child_fields:
@@ -1905,7 +1925,9 @@ object MilvusBackfill {
     */
   private def extractMetadataFromSnapshot(
       metadata: SnapshotMetadata,
-      v2Segments: Seq[com.zilliz.spark.connector.read.V2SegmentInfo] = Seq.empty
+      v2Segments: Seq[com.zilliz.spark.connector.read.V2SegmentInfo] = Seq.empty,
+      endpoint: String = "",
+      writeBucket: String = ""
   ): (Long, Map[Long, Long], Map[Long, String]) = {
     val collectionID = metadata.snapshotInfo.collectionId
 
@@ -1918,23 +1940,48 @@ object MilvusBackfill {
       val segId = item.segmentID
       MilvusSnapshotReader.parseManifestContent(item.manifest) match {
         case Right(mc) =>
+          // A qualified manifest basePath names the bucket its data lives in.
+          // The stripped key below becomes milvus.writer.customPath and is
+          // resolved against fs.bucket_name (config.s3BucketName), so writing
+          // to a different bucket would report success while Milvus (reading
+          // its own bucket) sees nothing. Reject the mismatch up front, like
+          // the read path's validateSnapshotBucketForRelativeDataPaths.
+          MilvusStoragePath
+            .bucketOf(mc.basePath, endpoint, writeBucket)
+            .filter(bucket => writeBucket.nonEmpty && bucket != writeBucket)
+            .foreach { manifestBucket =>
+              throw new IllegalArgumentException(
+                s"Snapshot manifest basePath '${mc.basePath}' names bucket " +
+                  s"'$manifestBucket' but backfill --s3-bucket is '$writeBucket'; " +
+                  "refusing to write addfield data into the wrong bucket"
+              )
+            }
+          // Reduce Milvus-format s3://<address>/<bucket>/<key> basePaths to the
+          // scheme-less bucket-relative key that the native writer consumes;
+          // it cannot read a scheme-bearing customPath for lack of an extfs.*
+          // config. This also keeps the insert_log slice below stable.
+          val basePath = MilvusStoragePath.toBucketRelativeKey(
+            mc.basePath,
+            endpoint,
+            writeBucket
+          )
           // Extract partition ID from basePath: .../insert_log/{col_id}/{part_id}/{seg_id}
-          val parts = mc.basePath.split("/")
+          val parts = basePath.split("/")
           val insertLogIdx = parts.indexOf("insert_log")
           if (insertLogIdx >= 0 && insertLogIdx + 2 < parts.length) {
             try {
               val partitionId = parts(insertLogIdx + 2).toLong
-              segmentBasePathMap += (segId -> mc.basePath)
+              segmentBasePathMap += (segId -> basePath)
               segmentToPartitionMap += (segId -> partitionId)
             } catch {
               case _: NumberFormatException =>
                 logger.warn(
-                  s"Skipping segment $segId: failed to parse partition ID from basePath: ${mc.basePath}"
+                  s"Skipping segment $segId: failed to parse partition ID from basePath: $basePath"
                 )
             }
           } else {
             logger.warn(
-              s"Skipping segment $segId: basePath does not contain expected insert_log structure: ${mc.basePath}"
+              s"Skipping segment $segId: basePath does not contain expected insert_log structure: $basePath"
             )
           }
         case Left(e) =>
@@ -1983,7 +2030,14 @@ object MilvusBackfill {
     // Same s3:// → s3a:// normalization as readSnapshotJson /
     // readBackfillData. Without it, an s3:// URL would route to the legacy
     // S3FileSystem which ignores fs.s3a.bucket.<b>.* config.
-    val outputPath = normalizeObjectStorageScheme(rawOutputPath, config)
+    val outputPath =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun"))
+        normalizeObjectStorageScheme(rawOutputPath, config)
+      else
+        // --output is a user-supplied path; passing the Milvus target bucket
+        // as the detection hint would misclassify a key whose first segment
+        // happens to equal it (see the source-path site above).
+        normalizeS3Scheme(rawOutputPath, config.s3Endpoint, "")
     try {
       withScopedHadoopStorage(spark, outputPath, config, isSource = false) {
         val hadoopPath = new Path(outputPath)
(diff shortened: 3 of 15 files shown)


src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1946,8 +1952,12 @@ class MilvusScan(
       val entries = manifestList.flatMap { item =>
         val (basePath, requestedReadVersion) =
           MilvusSnapshotReader.parseManifestContent(item.manifest) match {
-            case Right(content) => (content.basePath, content.ver.toLong)
-            case Left(_)        => (item.manifest, -1L)
+            case Right(content) =>
+              (
+                MilvusSnapshotReader.toStandardS3Path(content.basePath),
```
liliu-z: `MilvusSnapshotReader.toStandardS3Path(content.basePath)` here only affects `loadV3DeletePlanning`; `buildSnapshotPartitions` later reparses the manifest (line 2067) and passes the original `content.basePath` straight to `MilvusLoonPartitionReader`, so the executor never sees this canonicalized value. The native reader (`manifest_c.cpp:64` -> `FilesystemCache.get`) rejects any scheme-bearing path without a matching `extfs.*` block that `Properties.scala` never emits, so with `base_path=s3://minio:9000/bucket/files/insert_log/...` Hadoop-side delete planning succeeds but the scan fails during reader init. Pass a scheme-less bucket-relative key at the FFI boundary (keep `s3a://bucket/key` only for Hadoop APIs).

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -207,7 +207,8 @@ object V2SegmentLoader extends Logging {
   def resolvePath(path: String, bucket: String): String = {
     if (path == null) path
     else if (path.startsWith("s3a://")) path
-    else if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://")
+    else if (path.startsWith("s3://"))
+      MilvusSnapshotReader.toStandardS3Path(path)
```
liliu-z: `resolvePath` canonicalizes only the AVRO manifest and sample-footer reads, but `MilvusSegmentManifestReader.toV2SegmentInfo` still stores `sorted.map(_.logPath)` and `MilvusPackedV2PartitionReader` forwards those original paths to the native reader. For an AVRO entry containing `s3://minio:9000/bucket/files/insert_log/...`, planning succeeds but the executor read fails with the same missing-`extfs.*` error. Convert every returned column-group path to the scheme-less key the native filesystem expects.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1990,10 +1993,14 @@ object MilvusBackfill {
...
   ): Unit = {
     if (path == null) return
-    if (!(path.startsWith("s3://") || path.startsWith("s3a://"))) return
-
-    // Extract bucket name from s3(a)://bucket/key
-    val withoutScheme = path.stripPrefix("s3a://").stripPrefix("s3://")
+    // Canonicalize Milvus-format s3://<address>/<bucket>/<key> to
+    // s3a://<bucket>/<key> so the bucket (and thus the per-bucket Hadoop
+    // S3A config) is extracted correctly.
+    val canonical = MilvusSnapshotReader.toStandardS3Path(path)
```
liliu-z: `toStandardS3Path` swallows `URISyntaxException` and returns its input verbatim, but the branch decision now runs on the canonicalized string, so an `s3://...` that is not a legal `java.net.URI` (a space, an unescaped `%`, `[`/`]`) falls out of the S3 branch entirely, where the old `path.startsWith("s3://")` check would have applied per-bucket S3A config. The same pattern at `:2113` is worse: `scala.io.Source.fromFile(canonicalPath)` at `:2132` then surfaces a `FileNotFoundException` naming an S3 URI. Branch on the raw scheme and use the canonical form only for the value.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,65 @@ object MilvusSnapshotReader {
...
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch { case _: Exception => return trimmed }
+    Option(uri.getScheme).map(_.toLowerCase) match {
+      case Some("s3") | Some("s3a") =>
+        val authority = Option(uri.getAuthority).map(_.trim).filter(_.nonEmpty)
+        val rawPath = Option(uri.getPath).getOrElse("").stripPrefix("/")
+        if (authority.exists(_.indexOf(':') >= 0)) {
```
liliu-z: Milvus-format detection keys off a colon in the authority, so it only recognizes endpoints that carry an explicit port. This repo's own docs and tests configure port-less endpoints (`fs.address`/`s3Endpoint` of `s3.amazonaws.com`, `storage.googleapis.com`), and the native writer normalizes `https://host` addresses down to a bare host when no port is present, so a Milvus-format location like `s3://s3.amazonaws.com/milvus-bucket/file/...` takes the standard-S3 branch and yields `s3.amazonaws.com` as the bucket — the same wrong-bucket failure this PR is fixing. Detect the shape from the configured endpoint (available as `fs.address` at the MilvusScan call sites and `config.s3Endpoint` in backfill) rather than from the presence of a port.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -207,7 +207,8 @@ object V2SegmentLoader extends Logging {
   def resolvePath(path: String, bucket: String): String = {
     if (path == null) path
     else if (path.startsWith("s3a://")) path
```
liliu-z: `resolvePath` returns any `s3a://` path untouched, so an endpoint-prefixed `s3a://minio:9000/bucket/key` keeps `minio:9000` as the bucket instead of collapsing to `s3a://bucket/key`. The PR treats that shape as a real input elsewhere — its own new tests exercise `snapshotBucketsToConfigure` and `buildSnapshotHadoopConf` with `s3a://eric-spark-minio:9000/milvus-bucket/...`, and `toStandardS3Path` handles it — so the early return is an inconsistency worth closing. The same `s3a://` early return exists in MilvusStorageV3ManifestReader.scala:149, ListV2SegmentsApp.scala:275, and ReadSourceOnlyApp.scala:271-272; route all four through `toStandardS3Path`, which is already idempotent for standard `s3a://bucket/key`.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1862,23 +1862,26 @@ object MilvusBackfill {
       val segId = item.segmentID
       MilvusSnapshotReader.parseManifestContent(item.manifest) match {
         case Right(mc) =>
+          // Canonicalize Milvus-format s3://<address>/<bucket>/<key> basePaths
+          // so the insert_log slice and downstream path construction are stable.
+          val basePath = MilvusSnapshotReader.toStandardS3Path(mc.basePath)
```
liliu-z: The canonicalized basePath stored here is not just used for parsing — it becomes the native writer's base path, and the native layer cannot consume a scheme-bearing path any more than it could the original. `s3a://bucket/files/insert_log/...` is parsed by milvus-storage as address=`bucket`, bucket=`files`, and then rejected for lack of a matching `extfs.*` block, so backfill against a Milvus-format basePath still fails at write time. Same fix as the reader side: keep the scheme-qualified form for Hadoop APIs only and hand the native writer a scheme-less bucket-relative key.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -892,20 +893,24 @@ object MilvusScan extends Logging {
...
+        case Some(other) if other != "s3" && other != "s3a" =>
+          throw new IllegalArgumentException(
+            s"Unsupported snapshot s3_location scheme '$other': $trimmed"
+          )
+        case _ =>
+          // Canonicalize bucket-relative / standard / Milvus-format shapes so
+          // the URI authority always resolves to the bucket. Without this, a
+          // Milvus-format s3://<address>/<bucket>/<key> location would yield
+          // the storage endpoint host (e.g. "eric-spark-minio") as the bucket.
+          val canonical = MilvusSnapshotReader.toStandardS3Path(trimmed)
```
liliu-z: Canonicalizing before the authority extraction defeats the userinfo handling that follows it. For `s3a://user:pass@bucket/files/snapshots/1/metadata/2.json`, `java.net.URI.getAuthority` returns `user:pass@bucket`, whose colon makes `toStandardS3Path` classify the location as endpoint-prefixed and rewrite it to `s3a://files/snapshots/1/metadata/2.json`, so `snapshotBucket` now reports `files` instead of `bucket`. The `takeWhile(_ != '@')` fallback below can no longer see the original authority; either strip userinfo before canonicalizing or ignore userinfo when testing for the endpoint-prefixed shape.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -2059,22 +2111,29 @@ class MilvusScan(
           case Right(content) => (content.basePath, content.ver.toLong)
           case Left(_)        => (item.manifest, -1L)
         }
+      // milvus-storage's native FFI resolves scheme-less keys against the
+      // fs.* properties and rejects scheme-bearing paths for lack of an
+      // extfs.* block, so hand it the bare bucket-relative key.
+      val manifestPath = MilvusSnapshotReader.toBucketRelativeKey(
```
liliu-z: `toBucketRelativeKey` strips the bucket off a fully-qualified V3 manifest basePath, and the resulting key is handed to `MilvusStorageV3InputPartition` together with `milvusOption`, whose `fs.bucket_name` is what the native FFI resolves it against (`Properties.fromMilvusOption` reads only `fs.bucket_name`, defaulting to `a-bucket` when unset). Nothing checks that the stripped bucket matches: `validateSnapshotBucketForRelativeDataPaths` filters V3 base paths through `isBucketRelativeSnapshotLocation`, so a qualified `s3a://data-bucket/files/...` path is dropped from the check entirely. Where the pre-PR code handed the native reader a scheme-bearing path (a hard failure), the read now silently targets whichever bucket the connector is configured with. Extract the bucket and pin it on the partition options (the `canonicalMilvusOption` built at line 2097 but only used for the V2 partitions), or reject a mismatch before stripping.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,153 @@ object MilvusSnapshotReader {
...
+    * shapes fall back to the trimmed, lowercased input.
+    */
+  private[connector] def storageEndpointHost(endpoint: String): String = {
+    if (endpoint == null) return ""
+    val trimmed = endpoint.trim
+    if (trimmed.isEmpty) return ""
+    val uri =
+      try new URI(if (trimmed.contains("://")) trimmed else "//" + trimmed)
+      catch { case _: Exception => return trimmed.toLowerCase }
+    Option(uri.getHost).getOrElse(trimmed.toLowerCase)
```
liliu-z: `Option(uri.getHost).getOrElse(trimmed.toLowerCase)` returns `URI.getHost` as-is, but `URI.getHost` preserves case while `isEndpointPrefixed` lowercases the host it compares against. A port-less configured endpoint like `S3.INTERNAL.EXAMPLE` therefore fails the equality check and is misclassified as the bucket. Apply `.map(_.toLowerCase)` to the successful host result.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,153 @@ object MilvusSnapshotReader {
...
+    if (path == null) return null
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch { case _: Exception => return trimmed }
+    Option(uri.getScheme).map(_.toLowerCase) match {
+      case Some("s3") | Some("s3a") =>
+        val authority = Option(uri.getAuthority).map(_.trim).filter(_.nonEmpty)
+        val rawPath = Option(uri.getPath).getOrElse("").stripPrefix("/")
```
liliu-z: `uri.getPath` percent-decodes, so reconstructing the URI from it loses the escaping the old prefix-only scheme swap preserved: `s3://bucket/file/a%20b.json` comes back as `s3a://bucket/file/a b.json`, which then throws in `V2SegmentLoader.readAllBytes`'s `new URI(...)` and in `snapshotBucket`'s `new URI(canonical)`, and `%23` decodes to a `#` that Hadoop parses as a fragment, addressing a different key. Use `getRawPath` here and at the matching `toBucketRelativeKey` site (line 659); `getRawAuthority` has the same problem on line 716.

src/main/scala/tools/ListV2SegmentsApp.scala
```diff
@@ -272,8 +272,8 @@ object ListV2SegmentsApp {
     */
   private def resolvePath(path: String, bucket: String): String = {
     if (path == null) path
-    else if (path.startsWith("s3a://")) path
-    else if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://")
+    else if (path.startsWith("s3a://") || path.startsWith("s3://"))
+      MilvusSnapshotReader.toStandardS3Path(path)
```
liliu-z: The tool accepts `--s3-endpoint` but `MilvusSnapshotReader.toStandardS3Path(path)` canonicalizes with the empty default, so a port-less Milvus-format URI like `s3://storage.internal/bucket/key` treats `storage.internal` as the bucket and fails. Thread the parsed endpoint through the helper.

src/main/scala/tools/ReadSourceOnlyApp.scala
```diff
@@ -269,7 +269,9 @@ object ReadSourceOnlyApp {
   }
 
   private def normalizeS3(path: String): String =
-    if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://") else path
+    if (path.startsWith("s3://") || path.startsWith("s3a://"))
+      MilvusSnapshotReader.toStandardS3Path(path)
```
liliu-z: The tool accepts `--s3-endpoint` but `MilvusSnapshotReader.toStandardS3Path(path)` canonicalizes with the empty default, so a port-less Milvus-format URI like `s3://storage.internal/bucket/key` treats `storage.internal` as the bucket and fails. Thread the parsed endpoint through the helper.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -950,31 +966,51 @@ object MilvusScan extends Logging {
...
+      endpoint: String = ""
   ): Unit = {
-    snapshotBucket(snapshotPath).foreach { snapshot =>
+    snapshotBucket(snapshotPath, endpoint).foreach { snapshot =>
       val relativeStorageV3Paths = storageV2ManifestList
-        .map(storageV2ManifestBasePath)
+        .map(item => storageV2ManifestBasePath(item, endpoint))
         .filter(isBucketRelativeSnapshotLocation)
       val relativeStorageV2Paths = v2Segments
         .flatMap(_.columnGroups.flatMap(_.filePaths))
```
liliu-z, line 1064: `validateSnapshotBucketForRelativeDataPaths` only throws when data paths are *relative* and buckets disagree, but `v2Segments` reaching `.filter(isBucketRelativeSnapshotLocation)` have already been through `toNativeSegment`, so `isBucketRelativeSnapshotLocation` always holds and the filter carries no information. With snapshot and connector both in bucket `A` and V2 `filePaths` qualified to `s3a://B/files/...`, `connectorBucket.contains(snapshot)` is true and the stripped keys silently resolve against `A` instead of producing the intended 'Refusing to guess which bucket' error. Capture the bucket before `toNativeSegment` strips it.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1862,23 +1864,31 @@ object MilvusBackfill {
       val segId = item.segmentID
       MilvusSnapshotReader.parseManifestContent(item.manifest) match {
         case Right(mc) =>
+          // Reduce Milvus-format s3://<address>/<bucket>/<key> basePaths to the
+          // scheme-less bucket-relative key that the native writer consumes;
+          // it cannot read a scheme-bearing customPath for lack of an extfs.*
+          // config. This also keeps the insert_log slice below stable.
+          val basePath = MilvusSnapshotReader.toBucketRelativeKey(
```
liliu-z: The manifest basePath is reduced to a bucket-relative key and stored in `segmentBasePathMap`, which becomes `milvus.writer.customPath` while `fs.bucket_name` is set to `config.s3BucketName` — so the bucket the manifest actually names is discarded without ever being compared to the configured write bucket. Run with `--s3-bucket other-bucket` against a snapshot whose basePath is `s3://minio:9000/milvus-bucket/files/insert_log/...` and the addfield binlogs plus the new manifest version land in `other-bucket`; the job reports success while Milvus, reading `milvus-bucket`, sees no change. The read path refuses to guess in this situation (`validateSnapshotBucketForRelativeDataPaths`); the write path has no equivalent check.

src/main/scala/read/MilvusStorageV3ManifestReader.scala
```diff
@@ -146,8 +146,8 @@ object MilvusStorageV3ManifestReader {
       path: String
   ): String = {
     if (path == null || path.isEmpty) path
-    else if (path.startsWith("s3a://")) path
-    else if (path.startsWith("s3://")) "s3a://" + path.stripPrefix("s3://")
+    else if (path.startsWith("s3a://") || path.startsWith("s3://"))
+      MilvusSnapshotReader.toStandardS3Path(path)
```
liliu-z: `resolveManifestDeltaPath` canonicalizes manifest delta-log paths with the default empty endpoint, and neither it nor its callers (`projectDeltaLogs`, `loadDeltaLogs`) accept one. A delta-log path written in Milvus format against a port-less endpoint keeps the endpoint host as its bucket, so the delete-plan read targets a nonexistent bucket — the same gap the PR closed elsewhere by threading `endpoint` through.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,199 @@ object MilvusSnapshotReader {
...
+          // Milvus-format: scheme://address/bucket/key — the first path
+          // segment is the bucket.
+          val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+          val slash = rawPath.indexOf('/')
+          val bucket = if (slash < 0) rawPath else rawPath.substring(0, slash)
+          Option(bucket).filter(_.nonEmpty)
+        } else {
+          // Standard S3: scheme://bucket/key — the authority host is the bucket.
+          Option(uri.getHost).orElse(
+            Option(uri.getRawAuthority)
```
liliu-z: `Option(uri.getRawAuthority).map(_.takeWhile(_ != '@').split(":").head)` keeps the userinfo instead of the host. `s3a://user:pass@a_bucket/files/...` has `getHost == null` (underscore), so `takeWhile(_ != '@')` yields `user:pass` and `.split(":").head` yields `user`—the bucket becomes `user`. This was harmless when it only fed a display value in the old `snapshotBucket`, but now it feeds `withPinnedBucket` at `MilvusDataSource.scala:2105`/`:2255`, pinning a bogus `fs.bucket_name` onto the partition.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -871,51 +873,66 @@ object MilvusScan extends Logging {
...
+        // Canonicalize both standard s3://bucket/key and Milvus-format
+        // s3://<address>/<bucket>/<key> to s3a://bucket/key.
+        MilvusSnapshotReader.toStandardS3Path(trimmed, endpoint = endpoint)
       case Some(other) =>
         throw new IllegalArgumentException(
           s"Unsupported snapshot s3_location scheme '$other': $trimmed"
         )
       case None if bucket.trim.nonEmpty =>
-        s"s3a://${bucket.trim}/${trimmed.stripPrefix("/")}"
+        MilvusSnapshotReader.toStandardS3Path(trimmed, bucket)
```
liliu-z: The guard above (`case None if bucket.trim.nonEmpty`) trims `bucket`, but the call `MilvusSnapshotReader.toStandardS3Path(trimmed, bucket)` passes it raw, so `bucket = " a-bucket "` yields `s3a:// a-bucket /files/...` and a `URISyntaxException` at the next `new URI`. The replaced code used `s"s3a://${bucket.trim}/..."`. Latent today because callers pre-trim via `connectorS3BucketOption`, but trimming in the guard and not the use is the asymmetry that bites later.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1862,23 +1870,47 @@ object MilvusBackfill {
...
+          // resolved against fs.bucket_name (config.s3BucketName), so writing
+          // to a different bucket would report success while Milvus (reading
+          // its own bucket) sees nothing. Reject the mismatch up front, like
+          // the read path's validateSnapshotBucketForRelativeDataPaths.
+          MilvusSnapshotReader
+            .bucketOf(mc.basePath, endpoint)
+            .filter(bucket => writeBucket.nonEmpty && bucket != writeBucket)
+            .foreach { manifestBucket =>
+              throw new IllegalArgumentException(
+                s"Snapshot manifest basePath '$mc.basePath' names bucket " +
```
liliu-z: `s"Snapshot manifest basePath '$mc.basePath' names bucket "` interpolates `mc` and then appends the literal `.basePath`, so the bucket-mismatch error dumps the whole `ManifestContent` case class followed by `.basePath` instead of the actual path string. The guard still fires correctly, but the diagnostic is useless.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -101,7 +103,8 @@ object V2SegmentLoader extends Logging {
       entry: AvroManifestEntry,
       bucket: String,
       hadoopConf: Configuration,
-      applyDeletes: Boolean = true
+      applyDeletes: Boolean = true,
+      endpoint: String = ""
```
liliu-z, line 108: The `endpoint` parameter added to `buildV2SegmentInfoFromEntry` is never used in the body: the one `resolvePath` call inside it (line 176, the parquet-footer sample used to recover field ids) still passes only `(path, bucket)`, so it canonicalizes with `endpoint = ""`. Milvus-format paths whose authority carries a port still work (the port signal alone triggers detection), but with a port-less configured endpoint — `fs.address = s3.amazonaws.com` / `storage.googleapis.com`, exactly the case `isEndpointPrefixed`'s host-equality branch was added for — `s3://s3.amazonaws.com/milvus-bucket/...` is left as `s3a://s3.amazonaws.com/milvus-bucket/...` and the footer read fails with the endpoint host treated as the bucket, aborting the whole V2 segment load. The equivalent call in the tools copy (ListV2SegmentsApp.scala:162) does thread the endpoint, which makes this look like a missed call site rather than intent.

src/main/scala/read/V2SegmentLoader.scala
```diff
@@ -203,11 +206,20 @@ object V2SegmentLoader extends Logging {
...
+  /** Prefix `bucket` when `path` has no scheme; pass through s3a:// / s3://.
+    *
+    * `endpoint` is the configured storage endpoint (e.g. `fs.address` /
+    * `s3Endpoint`); it lets [[MilvusSnapshotReader.toStandardS3Path]]
+    * recognize port-less endpoint-prefixed Milvus-format URIs.
+    */
+  def resolvePath(
+      path: String,
+      bucket: String,
+      endpoint: String = ""
```
liliu-z, line 219: `resolvePath` gained an `endpoint` parameter defaulted to "", but the StorageV2 delete-log path is still resolved without it: `MilvusDeltaLogReader.loadDeletePlan` calls `V2SegmentLoader.resolvePath(log.logPath, bucket)` (MilvusDeltaLogReader.scala:137) on the raw AVRO paths that `V2SegmentInfo` deliberately preserves. For a port-less custom endpoint the port heuristic cannot fire, so a Milvus-format delta-log path is misparsed with the endpoint host as the bucket — the exact failure this PR set out to fix, left in one path. Thread the endpoint through `loadDeletePlan` / `loadPartitionScopedDeletePlans` like the other call sites.

src/main/scala/tools/ReadSourceOnlyApp.scala
```diff
@@ -84,7 +84,8 @@ object ReadSourceOnlyApp {
           metadata.manifestList,
           s3Bucket,
           hadoopConf,
-          manifestSchemaVersion = metadata.manifestSchemaVersion
+          manifestSchemaVersion = metadata.manifestSchemaVersion,
+          endpoint = s3Endpoint
```
liliu-z, line 90: Threading `endpoint` into the driver-side segment loader only fixes half of this tool: the DataSource options it builds use `milvus.s3.*` keys that no code reads, so the executor-side FFI still falls back to `localhost:9000` / `a-bucket` / `minioadmin`. With `--s3-endpoint minio.remote:9000` the snapshot and manifest reads now succeed but `df.count()` fails. The gap predates this PR, but since the PR is threading the endpoint through this file it is worth populating the options from `Properties.FsConfig.*` (fs.address, fs.bucket_name, credentials) in the same pass.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -461,9 +466,10 @@ object MilvusBackfill {
   ): Either[BackfillError, DataFrame] = {
     // Hadoop 3.4.1 has separate s3:// and s3a:// FileSystem implementations
     // and per-bucket fs.s3a.bucket.<b>.* config is only honored by
-    // S3AFileSystem. Force the s3a scheme so the credentials we just wrote
-    // actually take effect.
-    val path = normalizeS3Scheme(rawPath)
+    // S3AFileSystem. Force the s3a scheme (and canonicalize any Milvus-format
+    // endpoint-prefixed path) so the credentials we just wrote actually take
+    // effect.
+    val path = normalizeS3Scheme(rawPath, config.s3Endpoint)
```
liliu-z: The backfill *source* path is canonicalized with the main `config.s3Endpoint` even though the source bucket has its own `sourceS3Endpoint` override, which the very same function pair consults for credentials a few lines later. The port heuristic covers the common `host:port` case, so this only bites a port-less source endpoint, but for consistency both this call and the one in `configureHadoopS3ForPath` should use `config.sourceS3Endpoint.getOrElse(config.s3Endpoint)` when `isSource` is true.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,203 @@ object MilvusSnapshotReader {
...
+    *   - the authority carries a port (`host:port`): an S3 bucket name may only
+    *     contain lowercase letters, digits, `-` and `.`, so it can never carry
+    *     a port. `uri.getHost` / `uri.getPort` ignore userinfo, so
+    *     `s3a://user:pass@bucket/...` is correctly treated as standard.
+    *   - the authority host equals the configured storage endpoint host. This
+    *     covers port-less endpoints (e.g. `s3.amazonaws.com`) that the
+    *     port-only test would miss.
+    */
+  private def isEndpointPrefixed(uri: URI, endpoint: String): Boolean = {
+    if (uri.getPort > 0) {
```
liliu-z: `isEndpointPrefixed` relies on `uri.getPort > 0`, but for a registry-style authority like `milvus_minio:9000` (docker-compose service naming), `java.net.URI` won't parse it as server-based because the underscore violates RFC 2396 `domainlabel`, so `getHost()` is null and `getPort()` is -1. The endpoint signal misses, and `storageEndpointHost` hits the same null-`getHost` path and falls back to the whole string (~608), producing `s3a://milvus_minio:9000/milvus-bucket/...` — the exact bucket-mismatch this PR exists to fix. Derive host/port from `getRawAuthority` textually (drop through the last `@`, split on the last `:`) in both functions.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,203 @@ object MilvusSnapshotReader {
...
+      path: String,
+      fallbackBucket: String = "",
+      endpoint: String = ""
+  ): String = {
+    if (path == null) return null
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch { case _: Exception => return trimmed }
```
liliu-z: The `catch { case _: Exception => return trimmed }` silently drops the `s3://`→`s3a://` rewrite on any unparseable URI. The call sites this replaced (`V2SegmentLoader.resolvePath`, `MilvusBackfill.normalizeS3Scheme`) guaranteed `s3a://` unconditionally to guard against the legacy `s3://` FileSystem ignoring `fs.s3a.bucket.<b>.*`; now `--snapshot-path "s3://bucket/my snapshot/2.json"` returns the input verbatim and routes to the wrong FileSystem. Falling back to no canonicalization is fine; falling back to the wrong scheme is not — restore the prefix rewrite in the catch.

src/main/scala/read/MilvusSnapshotReader.scala
```diff
@@ -588,6 +589,203 @@ object MilvusSnapshotReader {
...
+          if (slash < 0) {
+            s"s3a://$rawPath"
+          } else {
+            val bucket = rawPath.substring(0, slash)
+            val key = rawPath.substring(slash + 1)
+            s"s3a://$bucket/$key"
+          }
+        } else {
+          // Standard S3: scheme://bucket/key (also normalize s3 -> s3a).
+          s"s3a://${authority.getOrElse("")}/$rawPath"
```
liliu-z: The standard-S3 branch rebuilds the URI from `getRawAuthority` + `getRawPath`, both of which stop at a literal `#` or `?`. A key containing either character is silently truncated instead of being passed through, which is a regression from the string-concatenation the replaced call sites used and fails as a wrong-object read rather than an error. Appending the raw query/fragment (or falling back to the prefix rewrite when either is present) would preserve the old behavior.

src/main/scala/MilvusOption.scala
```diff
@@ -475,8 +476,10 @@ case class MilvusS3Option(
 
   def getFilePath(path: String): Path = {
     if (notEmpty(s3FileSystemType)) {
-      if (path.startsWith("s3a://")) {
-        new Path(path)
+      if (path.startsWith("s3a://") || path.startsWith("s3://")) {
+        // Canonicalize Milvus-format s3://<address>/<bucket>/<key> so the
+        // endpoint is not treated as the bucket.
+        new Path(MilvusSnapshotReader.toStandardS3Path(path))
```
liliu-z: This is the only new toStandardS3Path call site that does not thread the configured endpoint, even though `s3Endpoint` is a field of the enclosing case class. A port-less Milvus-format URI (e.g. `s3://oss-cn-hangzhou.aliyuncs.com/bucket/files/...`) is therefore still parsed with the endpoint as the bucket — the case the endpoint parameter exists to catch. Pass `endpoint = s3Endpoint`.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1853,7 +1899,8 @@ class MilvusScan(
     MilvusScan
       .snapshotBucketsToConfigure(
         snapshotPath,
-        MilvusScan.connectorS3BucketOption(rawOptions).getOrElse("")
+        MilvusScan.connectorS3BucketOption(rawOptions).getOrElse(""),
+        MilvusScan.connectorS3EndpointOption(rawOptions).getOrElse("")
```
liliu-z, line 1956: `connectorS3EndpointOption` accepts `s3.endpoint` in addition to `fs.address`, and this PR routes every Milvus-format detection decision through it — including line 1903 here — but `buildSnapshotHadoopConf` still reads the actual S3A endpoint only from `fs.address` at line 1833. A user who configures just `s3.endpoint=minio:9000` gets snapshot URIs canonicalized as MinIO-format while S3A is left pointing at its default AWS endpoint, so the read still fails on connection or auth. Either resolve the Hadoop endpoint through `connectorS3EndpointOption` too (and normalize the matching `s3.*` credential aliases), or restrict detection to `fs.address` so one option family is authoritative.

src/main/scala/MilvusOption.scala
```diff
@@ -9,6 +9,7 @@ import org.apache.hadoop.fs.s3a.S3AFileSystem
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 import com.zilliz.spark.connector.MilvusConnectionException
+import com.zilliz.spark.connector.read.MilvusSnapshotReader
```
liliu-z: Importing `com.zilliz.spark.connector.read.MilvusSnapshotReader` puts the config layer downstream of the read layer for what are generic storage-path utilities. Extracting `toStandardS3Path`/`toBucketRelativeKey`/`bucketOf` into something neutral like `com.zilliz.spark.connector.MilvusStoragePath` would let both layers depend on it without the config module reaching into `read`.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -1783,14 +1863,18 @@ class MilvusScan(
...
     val endpoint =
-      MilvusScan.optionValue(rawOptions, Properties.FsConfig.FsAddress)
+      MilvusScan.connectorS3EndpointOption(rawOptions)
     val accessKey =
-      MilvusScan.optionValue(rawOptions, Properties.FsConfig.FsAccessKeyId)
+      MilvusScan.connectorS3AccessKeyOption(rawOptions)
     val secretKey =
-      MilvusScan.optionValue(rawOptions, Properties.FsConfig.FsAccessKeyValue)
+      MilvusScan.connectorS3SecretKeyOption(rawOptions)
     val useSsl =
```
liliu-z, line 1887: The new alias resolution covers endpoint, access key, secret key and SSL but stops short of path-style access: `pathStyle` still reads only `fs.s3a.path.style.access` and `fs.use_virtual_host`, so `s3.pathStyleAccess` never reaches `fs.s3a.path.style.access`. A MinIO user who configures the documented `s3.*` family now gets a working S3A endpoint but virtual-host addressing (S3A's default when the key is unset), so the snapshot JSON read resolves `bucket.minio:9000` and fails. Note `MilvusS3Option` defaults `s3.pathStyleAccess` to "true" and `MilvusBackfill.configureHadoopS3ForPath` hardcodes it to "true", so this reader is the odd one out. Resolve `MilvusOption.S3PathStyleAccess` alongside the other aliases and apply it globally and per bucket.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,276 @@
...
+    val (textHost, textPort) = authorityHostPort(rawAuthority)
+    val port = if (uri.getPort > 0) uri.getPort else textPort.getOrElse(-1)
+    if (port > 0) {
+      true
+    } else {
+      val host = Option(uri.getHost)
+        .map(_.toLowerCase)
+        .orElse(Some(textHost.toLowerCase).filter(_.nonEmpty))
+        .getOrElse("")
+      host.nonEmpty && storageEndpointHost(endpoint) == host
```
liliu-z: For a port-less storage endpoint the only Milvus-format signal is an exact host-string match against the configured endpoint, so any spelling difference between what Milvus embedded in the URI and what the connector was given (regional vs global S3 host, IP vs service DNS name, an alias behind a 443 proxy) turns the endpoint host into the bucket. Downstream that surfaces badly: `bucketOf` returns the endpoint host, and the new backfill guard aborts with "names bucket '<endpoint-host>' but backfill --s3-bucket is ...", which points the user at the wrong thing. Consider a second signal that does not depend on endpoint spelling — first path segment equal to the configured bucket implies Milvus-format, authority equal to the configured bucket implies standard.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -2021,6 +2138,77 @@ class MilvusScan(
...
   }
 
+  private[sources] def withPinnedBucket(
+      milvusOption: MilvusOption,
+      bucket: String
+  ): MilvusOption = {
+    val trimmed = bucket.trim
+    if (trimmed.isEmpty) milvusOption
+    else
+      milvusOption.copy(
```
liliu-z: This PR makes the s3.* family a first-class alias for the driver (connectorS3EndpointOption/AccessKey/SecretKey/UseSsl feed buildSnapshotHadoopConf and Milvus-format detection), but the option map handed to native partitions is copied through unchanged except for fs.bucket_name, and Properties.fromMilvusOption reads only fs.* keys. A user who configures s3.endpoint/s3.accessKey/s3.secretKey now plans successfully on the driver and then fails on the executors, where the FFI falls back to localhost:9000 with minioadmin credentials. Translate the resolved endpoint/credentials/SSL into fs.address / fs.access_key_id / fs.access_key_value on the partition options (what ReadSourceOnlyApp is forced to do by hand) instead of only pinning the bucket.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,302 @@
...
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return trimmed
+    val uri =
+      try new URI(trimmed)
+      catch {
+        case _: Exception =>
+          // Unparseable s3 URI (a space, an unescaped %, [ ]). Restore the
+          // s3:// -> s3a:// prefix rewrite so the legacy s3:// FileSystem
+          // never swallows per-bucket S3A config.
+          if (trimmed.startsWith("s3://"))
```
liliu-z: The unparseable-URI fallback does a bare s3:// → s3a:// prefix rewrite, so an endpoint-prefixed path that java.net.URI rejects (space, brace, stray %) still comes back as `s3a://minio:9000/bucket/key`, leaving the endpoint as the S3A bucket — the exact failure this PR removes everywhere else. toBucketRelativeKey's fallback at line 155 is worse for the native reader: it returns a scheme-bearing s3a:// path where the FFI expects a bare key. Reachability is limited (the reader entry points call `new URI` first and throw, so only the backfill string-prefix entry points such as normalizeS3Scheme and readSnapshotJson reach it), but the fix is a purely textual split on the authority. The new test only covers the standard shape in this branch.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,353 @@
...
+      path: String,
+      endpoint: String = "",
+      configuredBucket: String = ""
+  ): Option[String] = {
+    if (path == null) return None
+    val trimmed = path.trim
+    if (trimmed.isEmpty) return None
+    val uri =
+      try new URI(trimmed)
+      catch { case _: Exception => return None }
```
liliu-z: bucketOf gives up and returns None for any URI java.net.URI rejects, while toBucketRelativeKey falls back to a textual split and still strips the bucket segment off the same input. nativeV3ManifestPath and nativeV2ColumnGroups call both, so for such a path the key is reduced to a bare object key but no bucket is pinned, and the native reader resolves it against the connector's fs.bucket_name — precisely the silent cross-bucket read the comment at MilvusDataSource.scala:2188-2192 says is prevented; the multi-bucket refusal in nativeV2ColumnGroups is blind to those paths for the same reason. Give bucketOf the same unparseable fallback the other two helpers got.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -2021,6 +2138,86 @@ class MilvusScan(
...
+    val trimmed = bucket.trim
+    val opts = milvusOption.options
+    var updated =
+      if (trimmed.isEmpty) opts else opts + (Properties.FsConfig.FsBucketName -> trimmed)
+    // The native FFI (Properties.fromMilvusOption) reads only fs.* keys, so a
+    // user who configured the s3.* aliases gets them translated here — without
+    // this the executors would fall back to localhost:9000 / minioadmin.
+    Seq(
+      (Properties.FsConfig.FsAddress, MilvusScan.connectorS3EndpointOption(opts)),
+      (Properties.FsConfig.FsAccessKeyId, MilvusScan.connectorS3AccessKeyOption(opts)),
```
liliu-z: The new s3.* -> fs.* translation copies s3.accessKey / s3.secretKey into fs.access_key_id / fs.access_key_value unconditionally, including when fs.use_iam=true. Under IAM the driver deliberately drops static credentials (buildSnapshotHadoopConf unsets fs.s3a.access.key/secret.key and installs the default provider chain), and Properties.fromMilvusOption only skips the minioadmin dev defaults — an fs.access_key_id that is actually present is always forwarded to the FFI. So an IRSA deployment with fs.use_iam=true plus a leftover s3.accessKey in the Spark options now gets a working driver and executors signing with stale static keys. Guard the AK/SK translation on the normalized fs.use_iam value.

src/main/scala/sources/MilvusDataSource.scala
```diff
@@ -871,51 +874,75 @@ object MilvusScan extends Logging {
 
     val scheme = Option(new URI(trimmed).getScheme).map(_.toLowerCase)
```
liliu-z: resolveClientSnapshotLocation and snapshotBucket still probe the scheme with a bare `new URI(trimmed)`, so a snapshot location java.net.URI rejects (a space, an unescaped %, brackets) throws a raw URISyntaxException at plan time instead of the descriptive IllegalArgumentException these functions otherwise raise. That also means the unparseable fallback that toStandardS3Path grew in this PR can never run for a client snapshot location, even though MilvusSnapshotReaderTest covers exactly that input shape. Probe the scheme with a helper that tolerates unparseable input (the same textual split MilvusStoragePath already uses) rather than removing the fallback, which is reachable from V2SegmentLoader.resolvePath, MilvusS3Option.getFilePath and normalizeS3Scheme.

src/main/scala/read/MilvusStorageV3ManifestReader.scala
```diff
@@ -19,15 +20,17 @@ object MilvusStorageV3ManifestReader {
...
-      hadoopConf: Configuration
+      hadoopConf: Configuration,
+      endpoint: String = ""
   ): Either[Throwable, Seq[V2DeltaLogFile]] = {
     try {
       val manifestPath = V2SegmentLoader.resolvePath(
         manifestFilePath(basePath, readVersion),
-        bucket
+        bucket,
+        endpoint
```
liliu-z: `endpoint` is passed positionally here (and again at :63) into `resolvePath`'s third parameter, which is `storageScheme`, not `endpoint`. With the common empty endpoint the path becomes `://bucket/_metadata/...` and `new URI` rejects it; with a MinIO endpoint it becomes `minio:9000://bucket/...`, an opaque URI whose scheme has no FileSystem. Since `loadV3DeletePlanning` runs whenever `applyDeletes` is on (default true) and the snapshot has a StorageV3 manifest list, every StorageV3 snapshot read fails. Pass `endpoint = endpoint` at both sites.

src/main/scala/read/MilvusDeltaLogReader.scala
```diff
@@ -128,13 +131,14 @@ object MilvusDeltaLogReader extends Logging {
...
-      hadoopConf: Configuration
+      hadoopConf: Configuration,
+      endpoint: String = ""
   ): Either[Throwable, MilvusDeletePlan] = {
     try {
       validatePkType(pkField)
       val plans = deltaLogs.map { log =>
         val fullyQualifiedPath =
-          V2SegmentLoader.resolvePath(log.logPath, bucket)
+          V2SegmentLoader.resolvePath(log.logPath, bucket, endpoint)
```
liliu-z: Same positional-argument mistake: `endpoint` lands in `resolvePath`'s `storageScheme` slot. `seg.deltaLogs` logPaths are already `s3a://bucket/key` (V2SegmentLoader qualified them), so with the default empty endpoint they are rewritten to `://bucket/key` and `readAllBytes` throws on `new URI`; with a MinIO endpoint they become `minio:9000://bucket/key`. Any V2 or V3 read that has delete logs fails. Use `endpoint = endpoint`.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -484,9 +490,19 @@ object MilvusBackfill {
...
-    val path = normalizeObjectStorageScheme(rawPath, config)
+    // S3AFileSystem. Force the s3a scheme (and canonicalize any Milvus-format
+    // endpoint-prefixed path, using the source bucket's endpoint override when
+    // present) so the credentials we just wrote actually take effect; Aliyun
+    // OSS paths keep their own scheme.
+    val path =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun"))
+        normalizeObjectStorageScheme(rawPath, config)
+      else
+        normalizeS3Scheme(
```
liliu-z, line 504: The source (`--parquet`) path is canonicalized with `configuredBucket = config.s3BucketName`, which is the Milvus target bucket, and `configureHadoopS3ForPath` does the same at :2281 even when `isSource = true`. The endpoint is threaded correctly (`sourceS3Endpoint` first), so only the third detection signal misfires — but it does: when the source key's first segment happens to equal the target bucket name, `isEndpointPrefixed` classifies a standard source URI as Milvus-format, so `s3://data-lake/milvus-bucket/backfill.parquet` with `--s3-bucket milvus-bucket` is rewritten to `s3a://milvus-bucket/backfill.parquet` and the source credentials are then installed under `fs.s3a.bucket.milvus-bucket.*`, overwriting the target bucket's. `BackfillConfig` has no source bucket option, so pass `""` as the bucket for source paths.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -1983,7 +2030,11 @@ object MilvusBackfill {
     // Same s3:// → s3a:// normalization as readSnapshotJson /
     // readBackfillData. Without it, an s3:// URL would route to the legacy
     // S3FileSystem which ignores fs.s3a.bucket.<b>.* config.
-    val outputPath = normalizeObjectStorageScheme(rawOutputPath, config)
+    val outputPath =
+      if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun"))
+        normalizeObjectStorageScheme(rawOutputPath, config)
+      else
+        normalizeS3Scheme(rawOutputPath, config.s3Endpoint, config.s3BucketName)
```
liliu-z: `writeResultJson` passes `config.s3BucketName` as the Milvus-format detection hint for the user-supplied `--output` path. `isEndpointPrefixed` treats "first path segment equals the configured bucket" as proof of the Milvus format, so `--output s3://results/main-bucket/backfill.json` with `--s3-bucket main-bucket` is rewritten to `s3a://main-bucket/backfill.json` and the result JSON is silently written to the wrong bucket. The analogous source-path site at :501-508 deliberately passes `""` for exactly this reason; do the same here (and for the non-source branch at :2287, which the output path also flows through).

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,401 @@
...
+        // Re-append a literal '?'/'#' so the key is not truncated.
+        val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+        val suffix =
+          Option(uri.getRawQuery).map("?" + _).getOrElse("") +
+            Option(uri.getRawFragment).map("#" + _).getOrElse("")
+        if (isEndpointPrefixed(uri, endpoint, configuredBucket)) {
+          // Milvus-format: scheme://address/bucket/key — the bucket segment
+          // precedes the object key.
+          val slash = rawPath.indexOf('/')
+          val key = if (slash < 0) rawPath else rawPath.substring(slash + 1)
```
liliu-z: `if (slash < 0) rawPath` should return `""` in the Milvus-format branch, since `rawPath`'s first segment is the bucket itself. `toBucketRelativeKey("s3://minio:9000/milvus-bucket")` yields `"milvus-bucket"`, so `nativeV3ManifestPath` pins that bucket and asks the FFI for an object of the same name inside it. No production caller reaches it today, but it is a one-character fix duplicated at :171.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,403 @@
...
+        true
+      } else {
+        val configured = configuredBucket.trim.toLowerCase
+        if (configured.nonEmpty && host.nonEmpty) {
+          val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+          val slash = rawPath.indexOf('/')
+          val firstSegment =
+            (if (slash < 0) rawPath else rawPath.substring(0, slash)).toLowerCase
+          if (host == configured) false // authority IS the bucket → standard
+          else firstSegment == configured // bucket in path → Milvus-format
```
liliu-z: The last-resort branch in isEndpointPrefixed treats any URI whose first path segment equals the configured bucket as Milvus-format, with no endpoint evidence at all, so a standard `s3://other-bucket/<configured-bucket>/...` path is rewritten to `s3a://<configured-bucket>/...` and read from the wrong bucket. With `configuredBucket=milvus-bucket` and an endpoint host that is not `archive`, `toStandardS3Path("s3://archive/milvus-bucket/key")` returns `s3a://milvus-bucket/key`. This is reachable wherever a real bucket name is threaded as the detection hint (resolveClientSnapshotLocation, snapshotBucket, MilvusBackfill.readSnapshotJson, MilvusS3Option.getFilePath); note the backfill source/output sites deliberately pass "" to dodge exactly this. It only bites when the object key's first segment happens to equal the configured bucket name, so it is a narrow tail case rather than a common failure, but the misclassification is silent.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -2326,15 +2406,29 @@ object MilvusBackfill {
...
+        snapshotPath.startsWith("s3a://") ||
+        snapshotPath.startsWith("oss://")
+      // Canonicalize bucket-relative / standard / Milvus-format
+      // (s3://<address>/<bucket>/<key>) snapshot locations to s3a://bucket/key;
+      // Aliyun OSS paths keep their own scheme.
+      val s3Path = if (isS3) {
+        if (snapshotPath.startsWith("oss://"))
+          normalizeObjectStorageScheme(snapshotPath, config)
+        else
+          MilvusStoragePath.toStandardS3Path(
```
liliu-z, line 2424: readSnapshotJson now decides the OSS-vs-S3 rewrite from the incoming path's scheme instead of the configured cloud provider, which drops the Aliyun handling PR #110 added. For an Aliyun deployment whose snapshot path arrives with an s3:// or s3a:// scheme (the normal case — Milvus writes s3:// locations regardless of the OSS backend), the path is now canonicalized to s3a:// rather than rewritten to oss://, so withScopedHadoopStorage takes the S3A branch, fs.oss.* credentials and the OSS security token are never applied, and spark.read.text hits the OSS endpoint through S3A. The two sibling call sites in this PR (readBackfillData at 494-508 and the output path at 2031-2040) gate on config.s3CloudProvider; this one should too. No test covers it.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,410 @@
...
+    val (authority, path, _, port) = unparseableAuthorityParts(trimmed)
+    if (port.nonEmpty) {
+      val slash = path.indexOf('/')
+      if (slash < 0) "s3a://" + path
+      else {
+        val bucket = path.substring(0, slash)
+        val key = path.substring(slash + 1)
+        s"s3a://$bucket/$key"
+      }
+    } else s"s3a://$authority/$path"
```
yhmo, line 218: src/main/scala/MilvusStoragePath.scala line:218
Low ---- The unparseable-s3 fallbacks (unparseableStandardS3Path, unparseableBucket, unparseableBucketRelativeKey) detect Milvus-format only via the port, so a port-less endpoint-prefixed URI that java.net.URI rejects (a space, a brace, a stray %) is canonicalized with the endpoint host kept as the bucket.

Example: toStandardS3Path("s3://storage.internal/milvus-bucket/my file.parquet", "", "storage.internal", "milvus-bucket") returns "s3a://storage.internal/milvus-bucket/my file.parquet", while the parseable equivalent collapses to "s3a://milvus-bucket/my file.parquet" via the endpoint-host / first-segment signals in isEndpointPrefixed. This is reachable from every toStandardS3Path call site (resolveClientSnapshotLocation, MilvusBackfill.readSnapshotJson, V2SegmentLoader.resolvePath, MilvusS3Option.getFilePath, ...).

Suggested fix: thread endpoint/configuredBucket into the unparseable helpers and apply the same textual detection used for parseable URIs, and add a test for the port-less unparseable combination (the new unparseable tests only exercise the port-bearing shape). This complements the port-bearing gap fixed in the current code (which the earlier thread on the toBucketRelativeKey fallback covered); the port-less variant is not covered anywhere.

src/main/scala/MilvusStoragePath.scala
```diff
@@ -0,0 +1,410 @@
...
+        true
+      } else {
+        val configured = configuredBucket.trim.toLowerCase
+        if (configured.nonEmpty && host.nonEmpty) {
+          val rawPath = Option(uri.getRawPath).getOrElse("").stripPrefix("/")
+          val slash = rawPath.indexOf('/')
+          val firstSegment =
+            (if (slash < 0) rawPath else rawPath.substring(0, slash)).toLowerCase
+          if (host == configured) false // authority IS the bucket → standard
+          else if (host.contains("."))
```
yhmo, line 138: src/main/scala/MilvusStoragePath.scala line:138
Medium ---- The dotted-authority guard still cannot distinguish a valid dotted S3 bucket from a portless endpoint: with configuredBucket=milvus-bucket, toStandardS3Path("s3://external.example.com/milvus-bucket/files/data.parquet") returns "s3a://milvus-bucket/files/data.parquet", silently dropping bucket "external.example.com". This is a narrower reopen of the earlier single-label concern: host.contains(".") does not make the authority an endpoint, and the previous resolver preserved authority-as-bucket. Please avoid the first-segment fallback without endpoint evidence (or provide an explicit disambiguation contract) and add a dotted-standard-bucket regression test.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -2326,15 +2406,31 @@ object MilvusBackfill {
...
+      // to scala.io.Source.fromFile. Only the read value is canonicalized.
+      val isS3 = snapshotPath.startsWith("s3://") ||
+        snapshotPath.startsWith("s3a://") ||
+        snapshotPath.startsWith("oss://")
+      // Canonicalize bucket-relative / standard / Milvus-format
+      // (s3://<address>/<bucket>/<key>) snapshot locations to s3a://bucket/key.
+      // Aliyun rewrites any s3:// / s3a:// location to oss:// (Milvus writes
+      // s3:// regardless of the backend), gated on the configured provider like
+      // readBackfillData / writeResultJson.
+      val s3Path = if (isS3) {
```
yhmo, line 2420: src/main/scala/operations/backfill/MilvusBackfill.scala line:2420
Medium ---- The comment says bucket-relative, standard, and Milvus-format snapshot locations are canonicalized, but isS3 is false for "files/snapshots/foo.json", so s3Path remains bucket-relative and the else branch passes it to scala.io.Source.fromFile. With config.s3BucketName configured, this should resolve through toStandardS3Path with fallbackBucket=config.s3BucketName (as the client snapshot path already does); otherwise the three-shape contract documented here is not implemented. Please add a bucket-relative snapshot test as well.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -2326,15 +2406,31 @@ object MilvusBackfill {
...
+        if (config.s3CloudProvider.trim.equalsIgnoreCase("aliyun"))
+          normalizeObjectStorageScheme(snapshotPath, config)
+        else
```
yhmo, line 2423: src/main/scala/operations/backfill/MilvusBackfill.scala line:2423
Medium ---- Restoring the Aliyun provider gate is not sufficient for a Milvus-format URI: normalizeObjectStorageScheme only changes s3:// to oss://, so "s3://oss-cn-hangzhou.aliyuncs.com/my-bucket/files/snapshot.json" becomes "oss://oss-cn-hangzhou.aliyuncs.com/my-bucket/files/snapshot.json" and Hadoop OSS still treats the endpoint host as the bucket. V2SegmentLoader.resolvePath already canonicalizes to bucket/key before applying the OSS scheme; please share that behavior for snapshot/source/output paths and add a custom-endpoint Aliyun test.

LawrenceTL92:
Closing in favour of a narrower follow-up; the requirements are captured in https://github.com/zilliztech/spark-milvus/issues/118.

The problem is real and worth fixing, but this PR has grown into a general path canonicalizer routed through nine production files, and the open review threads show the port-less detection cannot be made safe: the "first path segment equals the configured bucket" fallback rewrites legitimate standard URIs (`s3://other-bucket/<configured-bucket>/…`, `s3://external.example.com/<configured-bucket>/…`) to the wrong bucket, `--output` and source paths are canonicalized with the target bucket as the hint, `readSnapshotJson` now picks OSS vs S3 from the incoming scheme and drops the provider gate #110 added, and two reader call sites (`buildV2SegmentInfoFromEntry`, `loadDeletePlan`) still resolve without the endpoint. The branch has also diverged from main after the backup datasource work (#114/#115).

A follow-up that (a) detects Milvus-format only via a ported authority or an exact match on the configured endpoint host, (b) canonicalizes only Milvus-produced paths, and (c) keeps the provider-gated OSS handling would be much smaller and reviewable. Happy to review that when it's up.

PR #113, opened 2026-08-24
fix: standardize spark-milvus package publishing

tsunghanjacktsai:
## Summary

- rebase the package build onto the latest `main`, including the OSS and backup datasource changes
- build the pinned `milvus-storage` with Conan 2, Rust, and the required Linux AIO development package
- build AMD64 and ARM64 packages on architecture-native workers and reject an explicit `TARGETARCH` mismatch
- package the JNI libraries and all transitive shared libraries under `native/linux-x86_64` or `native/linux-aarch64`, then patch their RUNPATH for relocation
- publish the runnable assembly as the primary Maven JAR while preserving `com.zilliz:spark-connector_2.13:<branch>-<arch>-SNAPSHOT`
- make the snapshot repository and credentials file configurable so Jenkins can publish to the internal Nexus without embedding credentials or an internal URL in this public repository
- retain the existing `PUBLISH_TO_CENTRAL` build argument during the Jenkins migration to the repository-neutral `PUBLISH_MAVEN` flag

## Validation

- `GIT_BRANCH=main sbt -batch 'Test / compile'`
- `MAVEN_SNAPSHOT_REPOSITORY_URL=https://nexus.zilliz.cc/repository/maven-snapshots/ GIT_BRANCH=main sbt -batch 'show version' 'show publishTo' 'show Compile / packagedArtifacts' 'show publish / packagedArtifacts'`
  - resolved version: `main-arm64-SNAPSHOT` on the local ARM64 host
  - resolved repository: the supplied Nexus snapshot URL
  - primary published JAR: `spark-connector-assembly-main-arm64-SNAPSHOT.jar`
- `git diff --check`

The local Docker daemon is unavailable, so Jenkins still needs to execute the complete native Docker build on both AMD64 and ARM64 workers before merge.

## Follow-up

- pass the internal Nexus repository URL, Nexus credentials file, and target architecture from Jenkins
- update the `spark-data-service` native path checks and SHA-256 pins after both packages are published


LawrenceTL92 (approved):
Merged main into the branch (Dockerfile conflict with #116: the build step is now `sbt 'compile; Test/compile; IntegrationTest/compile; assembly'`).

Why I'm approving without a Jenkins run: the Dockerfile on main is already broken against the pinned milvus-storage submodule — `conanfile.py` requires Conan >= 2.0 while main installs 1.61, the Rust format bridge isn't built at all, and `NativeLibraryLoader` looks for `native/linux-x86_64/…` (plus `ossl-modules/`) while main copies flat `native/*.so`. This PR brings the Dockerfile in line with all three, so merging cannot make the release build worse than it is today, and it unblocks the Jenkins work.

Verified locally on the merged tree with `GIT_BRANCH=main sbt 'show version' 'show publishTo' 'show Compile/packageBin/artifact' 'show Compile/packagedArtifacts'`: version `main-arm64-SNAPSHOT`, repository resolves to `MAVEN_SNAPSHOT_REPOSITORY_URL` (Sonatype snapshots by default), and the primary `spark-connector` jar artifact maps to `spark-connector-assembly-main-arm64-SNAPSHOT.jar` (a real assembly build ran). Docker isn't available on my machine, so the native stages still need the AMD64 + ARM64 Jenkins runs the description calls for.

Two follow-ups for the Jenkins wiring, not blockers:
1. `Credentials(mavenCredentialsFile)` uses sbt's properties format and needs `realm=`, `host=`, `user=`, `password=` — the old parser accepted a file with only `host/user/password`, so the Jenkins-provided file must include `realm` (`Sonatype Nexus Repository Manager` for Nexus).
2. The `/root/.sbt` cache mount that used to smuggle the credentials file into the build is gone, and nothing in the Dockerfile provides `MAVEN_CREDENTIALS_FILE` now; `RUN --mount=type=secret,id=maven_credentials,target=/root/.sbt/sonatype_central_credentials` around the publish step (with `--secret id=maven_credentials,src=…` from Jenkins) would avoid baking credentials into a layer.

PR #115, opened 2026-09-03
fix: Support multi-file column groups in backup reads

congqixia:
milvus-storage#657 fixed BuildLoonColumnGroups to write per-file [0, rc_i) ranges instead of group-cumulative offsets, which the packed reader intersects against each file's own zero-based row groups and was silently truncating every file after the first. Bump the submodule to cb22dfa and drop the temporary guard in BackupMetaReader that rejected a column group spanning multiple binlog files.

Restore per-file tail row-count reads on their own thread pool (FooterReadPool stays separate from SegmentReadPool so a segment worker blocked on Future.get() can never starve its own tail reads), re-add MilvusParquetFooterReader.readRowCount, and update BackupMetaReaderTest to expect per-file counts instead of rejection.

Add MilvusStorageMultiFileGroupTest, a native-reader integration test that builds one column group spanning two files via MilvusStorageColumnGroups.createFromGroups and pins total rows == sum of per-file rows. Update the design doc and reference docs accordingly.

### DIFF
--- a/src/main/scala/read/BackupMetaReader.scala
+++ b/src/main/scala/read/BackupMetaReader.scala
@@ -50,20 +50,22 @@ import io.milvus.grpc.schema.{
   * separate `DestKey` under the backup dir and records only the source key in
   * the meta — so object paths are always reconstructed from `backupDir` plus
   * the collection/partition/group/segment/field/log IDs, never taken from
-  * `log_path`. Three gaps vs. a Milvus snapshot are closed here:
-  *   1. milvus-backup persists only `log_size` per binlog, not `entries_num`.
-  *      The per-file row count is recovered by reading each (single) parquet
-  *      footer's row count
-  *      ([MilvusParquetFooterReader.readFieldIdsAndRowCount]). 2. The AVRO
-  *      segment-info (and hence the slot -> real field ID mapping) is not
-  *      copied by the backup; the real field IDs are recovered from the **head
-  *      file** of each column group via that file's own parquet schema
-  *      ([MilvusParquetFooterReader.readFieldIdsAndRowCount]) — matching
-  *      V2SegmentLoader, which assumes all files in a group share the schema.
-  *      3. L0 delete-only segments are created by Milvus without a
-  *      `StorageVersion` (0/omitted), so they are handled before any
-  *      storage-version filtering.
-  *
+   * `log_path`. Three gaps vs. a Milvus snapshot are closed here:
+   *   1. milvus-backup persists only `log_size` per binlog, not `entries_num`.
+   *      Per-file row counts are recovered by reading each binlog's parquet
+   *      footer ([MilvusParquetFooterReader.readRowCount], with the head file's
+   *      footer read once for both field IDs and its row count via
+   *      [MilvusParquetFooterReader.readFieldIdsAndRowCount]).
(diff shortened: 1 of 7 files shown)


src/main/scala/read/BackupMetaReader.scala
```diff
@@ -636,10 +636,25 @@ object BackupMetaReader extends Logging {
...
+          } else Seq.empty
           V2ColumnGroup(
             fieldIds = headInfo.fieldIds,
             filePaths = nativePaths,
-            fileRowCounts = Seq(headInfo.rowCount),
+            // Per-file row counts (the packed reader derives per-file [0, rc_i)
+            // ranges); the BuildLoonColumnGroups fix (milvus-storage#657) made
+            // multi-file column groups safe, so a group may span several binlog
+            // files.
+            fileRowCounts = headInfo.rowCount +: tailRowCounts,
```
liliu-z, line 657: With the multi-file guard removed, a backup read against a pre-#657 `libmilvus-storage.so` silently drops every row after the first file of each column group; when all groups share the same file split the native cross-group row check at `cpp/src/reader.cpp:135-141` passes, so the scan returns a short DataFrame with no error and no log. The library is not version-locked to the jar: `stripPlatformPrefix` (`NativeLibraryLoader.java:214-224`) returns null for the flat `native/*.so` layout that `Makefile:88` and `Dockerfile:76-78` produce, so `loadFromResources()` extracts nothing and loading falls through to `System.loadLibrary` off `java.library.path`/`LD_LIBRARY_PATH`, while `make package`/`quick-build` rebuild the C++ only when the `.so` is absent. `BackupMetaReader.scala:670-684` compares footer metadata against `num_of_rows` only, and `MilvusPackedV2PartitionReader.next()` (`:341-343`) ends the scan on the first null batch without ever comparing delivered rows to `sum(fileRowCounts)`.

congqixia: Thanks for the thorough analysis — the concern is valid and I have addressed it with a delivered-row guard at the read path.

**The guard:** `MilvusPackedV2PartitionReader` now tracks the physical rows delivered by the packed reader (accumulated per exhausted batch, so unaffected by delete filtering) and, at EOF, verifies they equal `sum(fileRowCounts)` of the column groups. On a mismatch it throws `IllegalStateException` instead of ending the scan on the first null batch, so a pre-#657 library that silently drops every file after the first in a multi-file group is refused loudly rather than returning a short DataFrame.

**Why physical rows, not delivered rows:** delete filtering skips rows in `next()` but never removes them from a batch, so the physical counter equals what the native reader actually produced. Only the pre-fix truncation makes it disagree with the declared per-file row counts.

**Scope:** the check lives in the shared packed-V2 reader, so it covers the snapshot, backfill, and backup read paths alike.

I agree the library is not version-locked to the jar and that the flat `native/*.so` layout can fall through to `System.loadLibrary` off `java.library.path`/`LD_LIBRARY_PATH`; pinning the native artifact is the durable fix and can follow up separately. The row-count guard closes the silent-short-read window now. Both the new `MilvusStorageMultiFileGroupTest` and the existing suites pass.

Added in commit `2c3f6cc`.

liliu-z (commented):
/lgtm
/approve

PR #116, opened 2026-09-04
chore: Split integration tests into src/it source set

congqixia:
Separate the tests that need a live Milvus server (:19530) and MinIO (:9000) from the fast unit tests so 'sbt test' can run in CI without external infrastructure.

- Move MilvusLoonWriterTest, MilvusDataSourceTest and MilvusBackfillTest to src/it/scala
- Add an 'it' configuration that extends Test and reuses the shared native/JVM settings (fork, javaOptions, envVars, unmanagedJars)
- Give integration tests a generous 600s timeout vs the 10s unit-test timeout; Defaults.itSettings is required so the config resolves src/it/* instead of inheriting Test's src/test dirs

Unit tests are unaffected and stay on 'sbt test'; integration tests run via 'sbt IntegrationTest/test'.


build.sbt
```diff
@@ -101,15 +127,21 @@ lazy val root = (project in file("."))
...
+      testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oDF", "-W", "10", "10")
+    )),
 
-    // Test timeout - 10 seconds per test to avoid hanging
-    Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oDF", "-W", "10", "10"),
+    // Integration tests: need Milvus server (:19530) + MinIO (:9000).
+    // Give them a generous timeout since they run against real infrastructure.
+    // Defaults.itSettings points the source/resource dirs at src/it/* (otherwise
+    // the config would inherit Test's src/test dirs via `extend Test`).
+    inConfig(IntegrationTest)(Defaults.itSettings ++ nativeTestSettings ++ Seq(
```
liliu-z, line 152: The only automated build entry points are `Makefile:125` (`sbt test`) and `Dockerfile:91` (`sbt 'compile; Test/compile; package'`); neither compiles nor runs the `it` config, and the repo has no `.github/`, Jenkinsfile, or CI config. Repo-wide, `IntegrationTest` appears only at build.sbt lines 90, 117, 140, and 142 — it is not referenced from the Makefile, the Dockerfile, or the README. `Test/compile` at `Dockerfile:91` was the only automated typecheck of these 1068 lines before this PR; now nothing typechecks them, so the three integration suites break silently on the next API refactor and stay broken until someone manually runs `sbt IntegrationTest/test`.

liliu-z:
**src/it/scala/loon/MilvusLoonWriterTest.scala:0** `MilvusLoonWriterTest` mixes two MinIO tests (lines 60, 107) with three tests that touch no network, Spark, or native code: lines 17 and 31 call only `MilvusLoonPartitionWriter.parsePositiveDoubleOption`, and line 39 relies on `MilvusLoonWriter.scala:218-223` running `parsePositiveDoubleOption` before `Properties.fromMilvusOption` at line 224, so the `"NaN"` input throws during argument validation. Moving the whole file to `src/it` drops all three from `sbt test`, including the `constructor validates variable-width density before bucket validation` regression test, and no remaining unit test covers that ordering.

build.sbt
```diff
@@ -101,15 +127,21 @@ lazy val root = (project in file("."))
...
+      testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oDF", "-W", "600", "600")
```
liliu-z: `IntegrationTest / testOptions` has no config-scoped definition — sbt puts `testOptions :== Nil` in GlobalScope (`Defaults.scala:1329-1338`) and `testTaskOptions` only defines task-scoped copies (`Defaults.scala:1402-1419`) — so the `+=` at line 143 delegates through `config("it") extend Test` (`Load.scala:215-218`) to `Test / testOptions` and yields `-oDF -W 10 10 -oDF -W 600 600`. `Defaults.createTestRunners` (`Defaults.scala:1577-1581`) concatenates both Arguments, and ScalaTest 3.2.15's `ArgsParser.parseSlowpokeConfig` reads `slowpokeArgs(1)`/`(2)` (`ArgsParser.scala:161-164`), so integration tests run with 10s, not 600s. Separately, `-W` is ScalaTest's slowpoke notifier — it only emits `AlertProvided` events (`Runner.scala:660-674`, `Framework.scala:682-683`) and never fails, cancels, or interrupts a test — so the "10s per test to avoid hanging" comment at line 133 and the "generous timeout" comment at line 139 both describe behavior that does not exist, leaving the two configs behaviorally identical apart from their source directories.

liliu-z:
**src/test/scala/example/HelloSpec.scala:0** `HelloSpec` stays in `src/test`, and its `get segment info` test builds a `MilvusClient` against `http://localhost:19530` and calls `getSegmentInfo`, which issues a real `httpClient.send` (`MilvusClient.scala:805-817`) using a client with a 10s connect timeout (`MilvusClient.scala:169-173`). The exception is swallowed into a `Failure` (`MilvusClient.scala:875-879`) and the test asserts nothing, so it stays green with or without a server — which is why the split missed it. This falsifies the "no external services" contract this PR introduces at `build.sbt:133`, and the hardcoded segment IDs mean the test can never detect a regression.

liliu-z:
**Dockerfile:0** `Dockerfile:91` runs `sbt 'compile; Test/compile; package'` and `Makefile:125` runs `sbt test`; neither touches the `it` config, and `Test/compile` does not trigger `it/compile` — the dependency runs the other way (`it extend Test`, `build.sbt:90`). The 912 lines moved into `src/it` (`MilvusDataSourceTest.scala`, `MilvusBackfillTest.scala`) were type-checked by every Docker image build before this PR and are now type-checked by nothing, so a production API change that breaks them merges green and is only discovered whenever someone next hand-runs the integration suite.

build.sbt
```diff
@@ -87,12 +87,38 @@ lazy val gitBranch = {
...
 
+lazy val IntegrationTest = config("it") extend Test
+
+// Shared JVM/native setup required by both the unit-test and integration-test
+// source sets (native library loading, Arrow JNI access, JNI thin JAR).
+lazy val nativeTestSettings: Seq[Setting[_]] = Seq(
+  fork := true,
+  parallelExecution := true,
+  logBuffered := false,
+  javaOptions ++= Seq(
```
liliu-z: `nativeTestSettings` is applied under both `inConfig(Test)` and `inConfig(IntegrationTest)`, and `javaOptions` has no config-scoped definition in sbt, so `it / javaOptions ++=` delegates to the already-populated `Test / javaOptions` and appends the same list again — `sbt "show IntegrationTest/javaOptions"` returns 20 entries against 10 for `Test`. This is the exact delegation behaviour the comment at build.sbt:141-145 documents for `testOptions` and fixes there with `:=`. Today every duplicated flag is idempotent, but any future `Test`-scoped JVM flag lands in the integration-test JVM twice.

LawrenceTL92 (approved):
Verified locally on the PR head (f649ff7):

- `sbt "show IntegrationTest/javaOptions"` → 10 entries (same as `Test`), no duplication; `IntegrationTest/testOptions` → only `-W 600 600`; `IntegrationTest/scalaSource` → `src/it/scala`.
- `Test/compile` and `IntegrationTest/compile` both succeed; Dockerfile/Makefile now type-check the `it` config.
- `HelloSpec` no longer opens a connection to :19530.
- `sbt test`: 515/517 pass without Milvus/MinIO. The 2 failures (`MilvusStorageFFITest`, `MilvusStorageMultiFileGroupTest`) are only because the native `.so` isn't built on my machine; both use `fs.storage_type=local`, so keeping them in `src/test` is correct.

All three open threads are addressed by the latest commit. LGTM.

PR #117, opened 2026-09-08
enhance: Add input format dispatch to backfill reads

congqixia:
Backfill input reads are hardcoded to spark.read.parquet inside readBackfillData, so the CLI and config expose no way to select a different input source.

Introduce an inputFormat field on BackfillConfig, dispatch readBackfillData by format, and extract the existing parquet logic into readParquet. Iceberg and lance pass config validation but are rejected with an explicit unsupported-format error until their readers land.

The CLI gains --input-format and --input-path; --parquet remains the legacy alias. Supplying both paths fails fast instead of silently picking one.

### DIFF
--- a/src/main/scala/operations/backfill/BackfillApp.scala
+++ b/src/main/scala/operations/backfill/BackfillApp.scala
@@ -13,7 +13,12 @@ import com.zilliz.spark.connector.MilvusOption
   * --s3-secret-key <secret> \ [--s3-root-path <path>] [--s3-region <region>]
   * [--s3-cloud-provider aws|aliyun|gcp|azure|tencent|huawei] [--s3-use-ssl] \
   * [--batch-size <n>] [--output-result <path>] \ [--mode
-  * replace|coalesce|overwrite]
+  * replace|coalesce|overwrite] [--input-format parquet|iceberg|lance]
+  *
+  * --input-path <path> is a format-neutral alias for --parquet (mutually
+  * exclusive with it). --input-format selects the input reader; only
+  * "parquet" is implemented, iceberg/lance are validated but rejected until
+  * their readers land.
   *
   * --mode:
   *   - replace: parquet is absolute source of truth; unmatched source rows get
@@ -28,10 +33,9 @@ object BackfillApp {
   def main(args: Array[String]): Unit = {
     val parsed = parseArgs(args)
 
-    val parquetPath = parsed.getOrElse(
-      "parquet",
-      throw new IllegalArgumentException("--parquet is required")
-    )
+    val parquetPath = resolveInputPath(parsed)
+    val inputFormat =
+      parsed.getOrElse("input-format", BackfillConfig.DefaultInputFormat)
     val snapshotPath = parsed.getOrElse(
       "snapshot",
       throw new IllegalArgumentException("--snapshot is required")
@@ -85,6 +89,7 @@ object BackfillApp {
       sourceS3Region = parsed.get("source-s3-region"),
       batchSize = parsed.getOrElse("batch-size", "1024").toInt,
       columnMapping = parsed.get("column-mapping").map(parseColumnMapping),
+      inputFormat = inputFormat,
       mode = parsed.getOrElse("mode", MilvusOption.BackfillModeCoalesce)
     )
 
@@ -147,6 +152,8 @@ object BackfillApp {
   // later inside the AWS default provider chain.
   private[backfill] val KvFlags: Set[String] = Set(
     "parquet",
+    "input-path",
+    "input-format",
     "snapshot",
     "s3-endpoint",
     "s3-bucket",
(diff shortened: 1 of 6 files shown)


src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -168,6 +174,11 @@ case class BackfillConfig(
       )
     } else if (normalizedRoleArn.isEmpty && hasRoleDetails) {
       Left("s3RoleSessionName and s3ExternalId require s3RoleArn")
+    } else if (!BackfillConfig.AllowedInputFormats.contains(inputFormat.trim)) {
```
liliu-z, line 176: validate() (BackfillConfig.scala:177) accepts all of `AllowedInputFormats` = {parquet, iceberg, lance}, but only parquet has a reader (MilvusBackfill.scala:489-499). `--input-format iceberg|lance` passes validation at MilvusBackfill.scala:107, reads the snapshot manifest (:124), materializes every StorageV2 segment through AVRO manifests and parquet footers (:136), and opens a Milvus client (:155) before failing at :200 — a full cluster job and two rounds of S3 I/O spent on an error that is fully determined at argument-parse time. `NativeAssumeRoleCloudProviders` (BackfillConfig.scala:164-174) is the in-file precedent for rejecting a legal-but-unimplemented value inside `validate()`.

src/main/scala/operations/backfill/MilvusBackfill.scala
```diff
@@ -475,12 +475,35 @@ object MilvusBackfill {
...
+    * remains as a defensive safety net for embedded callers that bypass
+    * validate().
     */
   private def readBackfillData(
       spark: SparkSession,
       rawPath: String,
       config: BackfillConfig
+  ): Either[BackfillError, BackfillSource] =
+    config.inputFormat.trim match {
+      case BackfillConfig.DefaultInputFormat =>
```
liliu-z: `readBackfillData` matches `case BackfillConfig.DefaultInputFormat` (MilvusBackfill.scala:491) and builds its fallback message from the same constant (:498), while `ImplementedInputFormats` (BackfillConfig.scala:436) independently declares which formats have a reader, with nothing linking the two. Changing the default to another format routes that format into `readParquet` and pushes `parquet` into the unsupported branch — no compile error, and both new tests iterate the same sets so they stay green. Adding a format to `ImplementedInputFormats` alone, which is exactly what the comment at BackfillConfig.scala:435 instructs, makes `validate()` pass and defers the failure to MilvusBackfill.scala:200, after `loadSnapshotMetadata` (:124) and `loadV2Segments` (:136) have already spent S3 I/O — the late failure commit `2b57ba6` was written to remove.

src/main/scala/operations/backfill/BackfillConfig.scala
```diff
@@ -71,6 +71,12 @@ case class BackfillConfig(
     sourceS3UseIam: Option[Boolean] = None,
     sourceS3Region: Option[String] = None,
 
+    // Input data source format. "parquet" is the only implemented reader;
```
liliu-z, line 74: BackfillConfig.scala:75-77 says iceberg and lance are "recognized by validation so configs can be authored ahead of their readers, but readBackfillData rejects them", while `validate()` at :182-189 rejects them itself — such a config can never run, and the rejecting code is not `readBackfillData`. MilvusBackfill.scala:481-483 justifies the fallback branch as a "safety net for embedded callers that bypass validate()", but `readBackfillData` is `private` (:485) with one caller at :200, reachable only through `run()`, which validates unconditionally at :107 — that caller cannot exist, making `case other` unreachable.

liliu-z:
**src/main/scala/operations/backfill/README.md:0** The backfill README documents every other CLI flag but neither new one: the spark-submit example at :33 still shows only `--parquet`, the Required-flags table at :84 still lists `--parquet` as unconditionally required although it is now mutually exclusive with `--input-path` (BackfillApp.scala:182-186), and the Optional-flags table at :89-97 has no `--input-format` row. The CLI surface is this PR's only user-visible change, and this file is updated alongside CLI changes elsewhere in the repo (79e1074 / #110).

src/main/scala/operations/backfill/BackfillApp.scala
```diff
@@ -167,6 +174,26 @@ object BackfillApp {
 
   private[backfill] val KnownFlags: Set[String] = BoolFlags ++ KvFlags
 
+  /** Resolve the input data path from `--parquet` (legacy) or `--input-path`
+    * (format-neutral alias). Supplying both is a config mistake, not an
+    * override, so it fails fast instead of silently picking one.
+    */
+  private[backfill] def resolveInputPath(parsed: Map[String, String]): String = {
+    if (parsed.contains("parquet") && parsed.contains("input-path")) {
```
liliu-z: `resolveInputPath` tests presence with bare `parsed.contains` (BackfillApp.scala:182), and `parseArgs` stores empty values verbatim (`--parquet ""` becomes `"parquet" -> ""`, :252-261). A templated spark-submit that passes `--parquet "$LEGACY" --input-path "$INPUT"` during migration aborts with "--parquet and --input-path are mutually exclusive" even though only `--input-path` carries a value; the same bare lookup lets `--input-path ""` resolve to `""` and reach `spark.read.parquet("")` at MilvusBackfill.scala:200, after snapshot and segment I/O. The repo's only comparable check does the opposite and documents why (`nonEmptyOption`, MilvusOption.scala:181-186 and the comment at :195-198).

LawrenceTL92 (approved):
Latest commits address the three open threads: `ImplementedInputFormats` is now the single source of truth for both `validate()` and `readBackfillData`, the config comment matches what validate() actually does, and `resolveInputPath` treats empty flag values as absent (with tests). I pushed a small follow-up commit documenting `--input-path` / `--input-format` in the backfill README, which was the remaining gap.

Verified locally on top of current main (after #116): `testOnly com.zilliz.spark.connector.operations.backfill.*` → 142/142 pass. LGTM.

PR #124, opened 2026-09-10
fix: harden native Docker build dependencies

tsunghanjacktsai:
## Summary

- install the OpenSSL, pkg-config, and libclang dependencies required by the pinned milvus-storage Rust/C++ build
- fail fast when OpenSSL headers, pkg-config metadata, or libclang are unavailable
- keep the pinned Conan recipes and checksums while replacing obsolete Avro and Boost download endpoints
- preserve the existing architecture-native AMD64/ARM64 packaging flow

## Validation

- full cold AMD64 Docker build in the documented remote build environment with `PUBLISH_MAVEN=false`
- output image: `linux/amd64`
- output artifact: `spark-connector-assembly-pr124-validation-amd64-SNAPSHOT.jar` (432 MB)
- verified `native/linux-x86_64/libmilvus-storage.so` and `libmilvus-storage-jni.so`
- verified all 256 bundled ELF shared libraries are x86_64
- verified no ARM64 native entries are present in the AMD64 artifact
- verified all bundled native dependencies resolve after installing `libaio1`, which spark-data-service already installs and checks at runtime
- GitHub Build & unit tests passed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates wget curl git g++ gcc make ccache gdb \
     python3 python3-pip \
     zip unzip \
-    automake autoconf libtool patchelf libaio-dev \
+    automake autoconf libtool patchelf libaio-dev libssl-dev pkg-config \
     && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


Thor-ChenBiao (approved):
LGTM

